PR #5, opened 2020-01-30
[FLINK-15769] [kafka-io] Add configuring startup position for Kafka ingress

tzulitai:
With this PR, the following options are now supported:
* from committed consumer group offsets in Kafka
* from earliest offset
* from latest offset
* from specific offsets
* from offsets with ingestion timestamp >= specified date

The changes simply expose configuration. Actual functionality is already implemented by the `FlinkKafkaConsumer`. The exposed configuration API is the following:
```
KafkaIngressBuilder<T> builder = KafkaIngressBuilder.forIdentifier(...)
    .withStartupPosition(KafkaIngressStartupPosition.fromGroupOffsets())
    .withStartupPosition(KafkaIngressStartupPosition.fromEarliest())
    .withStartupPosition(KafkaIngressStartupPosition.fromLatest())
    .withStartupPosition(KafkaIngressStartupPosition.fromSpecificOffsets(Map<KafkaTopicPartition, Long>))
    .withStartupPosition(KafkaIngressStartupPosition.fromDate(java.util.Date))
```

---

Apart from the main changes, there are 3 additional preliminary changes:
* 7e8e522 Add named method for setting consumer group id `withConsumerGroupId(String)` to builder
* 8d09c39 Add named method for setting auto offset reset position
* 0982eec Move Kafka properties resolution from `KafkaSourceProvider` to builder

The additional named methods makes sense, since they have an interplay with the startup position configs. For example, for `KafkaIngressStartupPosition.fromGroupOffsets()`, a consumer group id must be set.
This is not provided as part of the `fromGroupOffsets` signature because consumer group ids have functionality beyond startup position (e.g., we may later introduce offset committing back to Kafka which also relies on a consumer group id being set).

### DIFF
--- a/statefun-docs/src/main/java/org/apache/flink/statefun/docs/io/kafka/IngressSpecs.java
+++ b/statefun-docs/src/main/java/org/apache/flink/statefun/docs/io/kafka/IngressSpecs.java
@@ -21,7 +21,7 @@ import org.apache.flink.statefun.docs.models.User;
 import org.apache.flink.statefun.sdk.io.IngressIdentifier;
 import org.apache.flink.statefun.sdk.io.IngressSpec;
 import org.apache.flink.statefun.sdk.kafka.KafkaIngressBuilder;
-import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.flink.statefun.sdk.kafka.KafkaIngressStartupPosition;
 
 public class IngressSpecs {
 
@@ -31,8 +31,9 @@ public class IngressSpecs {
   public static final IngressSpec<User> kafkaIngress =
       KafkaIngressBuilder.forIdentifier(ID)
           .withKafkaAddress("localhost:9092")
+          .withConsumerGroupId("greetings")
           .withTopic("my-topic")
           .withDeserializer(UserDeserializer.class)
-          .withProperty(ConsumerConfig.GROUP_ID_CONFIG, "greetings")
+          .withStartupPosition(KafkaIngressStartupPosition.fromLatest())
           .build();
 }
--- a/statefun-flink/statefun-flink-io-bundle/src/main/java/org/apache/flink/statefun/flink/io/kafka/KafkaSourceProvider.java
+++ b/statefun-flink/statefun-flink-io-bundle/src/main/java/org/apache/flink/statefun/flink/io/kafka/KafkaSourceProvider.java
@@ -68,7 +68,7 @@ public class KafkaSourceProvider implements SourceProvider {
           convertKafkaTopicPartitionMap(offsetsPosition.getSpecificOffsets()));
     } else if (startupPosition.isDate()) {
       KafkaIngressStartupPosition.DatePosition datePosition = startupPosition.asDate();
-      consumer.setStartFromTimestamp(datePosition.getTime());
+      consumer.setStartFromTimestamp(datePosition.getEpochMilli());
     } else {
       throw new IllegalStateException("Safe guard; should not occur");
     }
--- a/statefun-flink/statefun-flink-io-bundle/src/main/java/org/apache/flink/statefun/flink/io/kafka/KafkaSourceProvider.java
+++ b/statefun-flink/statefun-flink-io-bundle/src/main/java/org/apache/flink/statefun/flink/io/kafka/KafkaSourceProvider.java
@@ -52,6 +52,8 @@ public class KafkaSourceProvider implements SourceProvider {
     Properties properties = new Properties();
     properties.putAll(spec.properties());
     properties.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, spec.kafkaAddress());
+    properties.put(
+        ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, spec.autoResetPosition().asKafkaConfig());
     spec.consumerGroupId()
         .ifPresent(groupId -> properties.put(ConsumerConfig.GROUP_ID_CONFIG, groupId));
 
--- a/statefun-flink/statefun-flink-io-bundle/src/main/java/org/apache/flink/statefun/flink/io/kafka/KafkaSourceProvider.java
+++ b/statefun-flink/statefun-flink-io-bundle/src/main/java/org/apache/flink/statefun/flink/io/kafka/KafkaSourceProvider.java
@@ -96,6 +96,6 @@ public class KafkaSourceProvider implements SourceProvider {
   private static org.apache.flink.streaming.connectors.kafka.internals.KafkaTopicPartition
       convertKafkaTopicPartition(KafkaTopicPartition partition) {
     return new org.apache.flink.streaming.connectors.kafka.internals.KafkaTopicPartition(
-        partition.getTopic(), partition.getPartition());
+        partition.topic(), partition.partition());
   }
 }
--- a/statefun-flink/statefun-flink-io-bundle/src/main/java/org/apache/flink/statefun/flink/io/kafka/KafkaSourceProvider.java
+++ b/statefun-flink/statefun-flink-io-bundle/src/main/java/org/apache/flink/statefun/flink/io/kafka/KafkaSourceProvider.java
@@ -17,7 +17,6 @@
  */
 package org.apache.flink.statefun.flink.io.kafka;
 
-import java.util.Properties;
 import org.apache.flink.statefun.flink.io.common.ReflectionUtil;
 import org.apache.flink.statefun.flink.io.spi.SourceProvider;
 import org.apache.flink.statefun.sdk.io.IngressSpec;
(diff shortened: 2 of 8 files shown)


statefun-kafka-io/src/main/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressBuilder.java
```diff
@@ -39,8 +39,7 @@
   private Class<? extends KafkaIngressDeserializer<T>> deserializerClass;
   private String kafkaAddress;
   private KafkaIngressAutoResetPosition autoResetPosition = KafkaIngressAutoResetPosition.LATEST;
-  private KafkaIngressStartupPosition startupPosition =
-      KafkaIngressStartupPosition.fromGroupOffsets();
+  private KafkaIngressStartupPosition startupPosition = KafkaIngressStartupPosition.fromLatest();
```
sjwiesman: Why latest? I think earliest is more intuitive (and also the default on the connector)

statefun-kafka-io/src/main/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressStartupPosition.java
```diff
@@ -0,0 +1,152 @@
...
+package org.apache.flink.statefun.sdk.kafka;
+
+import java.util.Date;
+import java.util.Map;
+
+/**
+ * Position for the ingress to start consuming Kafka partitions.
+ */
+@SuppressWarnings("WeakerAccess, unused")
+public class KafkaIngressStartupPosition {
```
sjwiesman: Can you add a private constructor to this class, that way it can only be extended by the static inner classes.

sjwiesman:
Can you add something to the docs about this? It can just be one line saying the starting positions are configurable and then a list of the options.

statefun-kafka-io/src/main/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressBuilder.java
tzulitai: The `FlinkKafkaConsumer` actually uses committed group offsets as the default:
https://github.com/apache/flink/blob/master/flink-connectors/flink-connector-kafka-base/src/main/java/org/apache/flink/streaming/connectors/kafka/FlinkKafkaConsumerBase.java#L159

The reason we chose to change this to latest, is because group offsets startup require that the user must set the consumer group id. We wanted to go with a default that requires the minimal setup to get started with using the ingress.

tzulitai: As for latest v.s. earliest, for a typical Kafka consumer (non-Flink/StateFun), when consuming from a topic that has no previous record of committed offsets, that's the default behaviour AFAIK.

See the section about `auto.offset.reset` in the Kafka docs here:
https://docs.confluent.io/current/clients/consumer.html#id1

statefun-kafka-io/src/main/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressStartupPosition.java
tzulitai: Good point! Will add this.

tzulitai:
> Can you add something to the docs about this? It can just be one line saying the starting positions are configurable and then a list of the options.

Will do!

statefun-kafka-io/src/main/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressBuilder.java
sjwiesman: Sounds good to me, I don't have a strong opinion on this one

sjwiesman:
I was going to ask if this should be configurable via YAML but I'm not really sure the state of those configurations and if that would be out of scope for this ticket.

tzulitai:
> I was going to ask if this should be configurable via YAML but I'm not really sure the state of those configurations and if that would be out of scope for this ticket.

@sjwiesman yes, I'll add YAML support as part of this PR as well (or maybe even a follow-up PR if this one gets a bit too big; they should be fairly independent code-wise).
Wanted to open the current API to have a preview of the behaviour and what to expose before moving on.

tzulitai:
Sorry @sjwiesman, I had to do a force push when addressing your comments because I accidentally squashed my fixup commits already 😅

tzulitai:
cc @igalshilman the PR is ready for review

statefun-kafka-io/src/main/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressAutoResetPosition.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.util.Locale;
+
+/** The auto offset reset position to use in case consumed offsets are invalid. */
+public enum KafkaIngressAutoResetPosition {
+  EARLIEST,
+  LATEST;
+
+  public String asKafkaConfig() {
+    return name().toLowerCase(Locale.ENGLISH);
```
igalshilman: I think that the actual mapping from a statefun enum name to a Kafka property name should happen within the SourceProvider, because this might be Kafka version specific.

statefun-kafka-io/src/main/java/org/apache/flink/statefun/sdk/kafka/KafkaTopicPartition.java
```diff
@@ -0,0 +1,73 @@
...
+    this.topic = Objects.requireNonNull(topic);
+
+    if (partition < 0) {
+      throw new IllegalArgumentException(
+          "Invalid partition id: " + partition + "; value must be larger or equal to 0.");
+    }
+    this.partition = partition;
+  }
+
+  public String getTopic() {
```
igalshilman: nit: can we stay consistent with the rest of the style in that package and omit `get` from the getters?

statefun-kafka-io/src/main/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressStartupPosition.java
```diff
@@ -0,0 +1,153 @@
...
+    public Map<KafkaTopicPartition, Long> getSpecificOffsets() {
+      return specificOffsets;
+    }
+  }
+
+  public static class DatePosition extends KafkaIngressStartupPosition {
+
+    private final Date date;
+
+    DatePosition(Date date) {
```
igalshilman: Do we need to use java.util.Date? it has some problems around mutability and thread-safety (I'm not sure it applies here) but can we either:
* pass in the long directly
* use a replacement for Date in java.util.time.* 
?

statefun-kafka-io/src/main/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressStartupPosition.java
```diff
@@ -0,0 +1,153 @@
...
+      this.date = date;
+    }
+
+    public long getTime() {
```
igalshilman: nit: same with the style

statefun-kafka-io/src/main/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressStartupPosition.java
```diff
@@ -0,0 +1,153 @@
...
+
+  /** Returns this position as a {@link DatePosition}. */
+  public DatePosition asDate() {
+    if (!isDate()) {
+      throw new IllegalStateException("This is not a startup position configured using a Date.");
+    }
+    return (DatePosition) this;
+  }
+
+  public static class GroupOffsetsPosition extends KafkaIngressStartupPosition {}
```
igalshilman: I think it would be better if all of these would be final, with a private constructor.

statefun-docs/src/main/java/org/apache/flink/statefun/docs/io/kafka/IngressSpecs.java
```diff
@@ -31,8 +31,9 @@
   public static final IngressSpec<User> kafkaIngress =
       KafkaIngressBuilder.forIdentifier(ID)
           .withKafkaAddress("localhost:9092")
+          .withConsumerGroupId("greetings")
```
igalshilman: I wondering rather we should set the withStartupPosition() in .withConsumerGroupId() in the ingress builder.

statefun-flink/statefun-flink-io-bundle/src/main/java/org/apache/flink/statefun/flink/io/kafka/KafkaSourceProvider.java
```diff
@@ -52,6 +52,8 @@
     Properties properties = new Properties();
     properties.putAll(spec.properties());
     properties.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, spec.kafkaAddress());
+    properties.put(
+        ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, spec.autoResetPosition().asKafkaConfig());
```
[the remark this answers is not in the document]
tzulitai: As discussed offline, I'll address this with a separate PR.

statefun-kafka-io/src/main/java/org/apache/flink/statefun/sdk/kafka/KafkaTopicPartition.java
tzulitai: will change

tzulitai:
Thanks for the review @igalshilman @sjwiesman.

I've addressed all comments except from the one made by Igal about configs being overwritten.
That one will be addressed as a follow-up PR.

Will proceed to clean up the fixup commits and merge this.

PR #6, opened 2020-01-31
[FLINK-15818] [kafka-io] Add startup postition configuration for Kafka ingress YAML support

tzulitai:
This is a follow-up to #5, and is based on that PR.
It extends the addition of configuring startup position / auto offset reset position / consumer group id to the YAML-based Kafka ingress.

Only the last 3 commits starting from c09fcd2 is relevant.

Moreover, this PR unifies the source creation logic between `ProtobufKafkaSourceProvider` and `KafkaSourceProvider`, to be always delegated through `KafkaSourceProvider`.
With this change, the `ProtobufKafkaSourceProvider` only deals with parsing the Json ingress spec into a `KafkaIngressSpec`, and delegates the actual creation of the Flink Kafka source to a `KafkaSourceProvider`.
This allows us to consolidate configuration validation logic in a single place (builder + `KafkaIngressSpec`).

Example YAML with the new configs:
```
ingress:
  meta:
    type: org.apache.flink.statefun.sdk.kafka/protobuf-kafka-connector
    id: com.mycomp.igal/names
  spec:
    address: kafka-broker:9092
    consumerGroupId: greeter
    topics:
      - names
    autoOffsetResetPosition: earliest
    startupPosition:
      type: specific-offsets
      offsets:
        - names/0: 91
        - names/1: 11
        - names/2: 8
    properties:
      - foo.config: bar
    messageType: org.apache.flink.test.SimpleMessage
    descriptorSet: classpath:test.desc
```

Other startup position types:
```
startupPosition:
      type: date
      date: 2020-01-31 16:01:49.440 +0800
```
```
startupPosition:
      type: earliest / latest / group-offsets
```

These can be validated using the existing `ProtobufKafkaSourceProviderTest#exampleUsage` test.

### DIFF
--- a/statefun-flink/statefun-flink-common/src/main/java/org/apache/flink/statefun/flink/common/json/Selectors.java
+++ b/statefun-flink/statefun-flink-common/src/main/java/org/apache/flink/statefun/flink/common/json/Selectors.java
@@ -33,6 +33,17 @@ public final class Selectors {
     return node.asText();
   }
 
+  public static Optional<String> optionalTextAt(JsonNode node, JsonPointer pointer) {
+    node = node.at(pointer);
+    if (node.isMissingNode()) {
+      return Optional.empty();
+    }
+    if (!node.isTextual()) {
+      throw new WrongTypeException(pointer, "not a string");
+    }
+    return Optional.of(node.asText());
+  }
+
   public static int integerAt(JsonNode node, JsonPointer pointer) {
     node = dereference(node, pointer);
     if (!node.isInt()) {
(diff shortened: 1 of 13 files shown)


tzulitai:
cc @igalshilman

statefun-flink/statefun-flink-common/src/main/java/org/apache/flink/statefun/flink/common/json/Selectors.java
```diff
@@ -89,6 +100,34 @@ public static int integerAt(JsonNode node, JsonPointer pointer) {
     return properties;
   }
 
+  public static Map<String, Long> longPropertiesAt(JsonNode node, JsonPointer pointer) {
```
igalshilman: Can you add a test for that in `SelectorsTest` ?

statefun-flink/statefun-flink-common/src/main/java/org/apache/flink/statefun/flink/common/json/Selectors.java
```diff
@@ -33,6 +33,17 @@ public static String textAt(JsonNode node, JsonPointer pointer) {
     return node.asText();
   }
 
+  public static Optional<String> optionalTextAt(JsonNode node, JsonPointer pointer) {
```
igalshilman: Can you add a test in `SelectorsTest`

statefun-flink/statefun-flink-io-bundle/src/main/java/org/apache/flink/statefun/flink/io/kafka/ProtobufKafkaSourceProvider.java
```diff
@@ -42,51 +54,175 @@
...
+  private final KafkaSourceProvider delegateProvider = new KafkaSourceProvider();
 
   @Override
   public <T> SourceFunction<T> forSpec(IngressSpec<T> spec) {
-    JsonNode json = asJsonIngressSpec(spec);
-    Properties properties = kafkaClientProperties(json);
-    List<String> topics = Selectors.textListAt(json, TOPICS_POINTER);
-    KafkaDeserializationSchema<T> deserializationSchema = deserializationSchema(json);
-    return new FlinkKafkaConsumer<>(topics, deserializationSchema, properties);
+    KafkaIngressSpec<T> kafkaIngressSpec = asKafkaIngressSpec(spec);
```
igalshilman: 👍 👍

statefun-flink/statefun-flink-common/src/main/java/org/apache/flink/statefun/flink/common/json/Selectors.java
tzulitai: Done 👌

statefun-flink/statefun-flink-common/src/main/java/org/apache/flink/statefun/flink/common/json/Selectors.java
tzulitai: Done 👌

tzulitai:
Thanks for the review @igalshilman.
I've addressed your comments, and will now proceed to merge this!

PR #7, opened 2020-02-01
[FLINK-15839] [kafka-io] Fix Stateful Functions KafkaIngressBuilder properties resolution logic

tzulitai:
This PR revisits how we resolve configs for the `KafkaIngressBuilder`, given that we have:
- Named configuration methods for important settings
- Some of which have default values, some don't
- Some eventually resolve as a Kafka client property, some are not
- At the same time, also accept a `Properties` which may also explicitly define values for the above.

The end result of the changes in this PR is as follows:
- Configs passed via named methods should always overwrite the value set via passed in `Properties`
- Any default values for named configuration methods should be defined in the builder
- If no config was passed via its named method, then we use the default value (if any) to overwrite the properties IFF the user also did not provide a value for it there.

---

The individual commits and their messages / descriptions should be sufficient in explaining what was changed to achieve this.

---

New tests in `KafkaIngressBuilderTest` was added to verify the changes.
Please also see there to get an overview of the expected behaviour.

### DIFF
--- a/statefun-kafka-io/src/main/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressBuilder.java
+++ b/statefun-kafka-io/src/main/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressBuilder.java
@@ -38,11 +38,15 @@ public final class KafkaIngressBuilder<T> {
   private final IngressIdentifier<T> id;
   private final List<String> topics = new ArrayList<>();
   private final Properties properties = new Properties();
-  private String consumerGroupId;
-  private KafkaIngressDeserializer<T> deserializer;
-  private String kafkaAddress;
-  private KafkaIngressAutoResetPosition autoResetPosition = KafkaIngressAutoResetPosition.LATEST;
-  private KafkaIngressStartupPosition startupPosition = KafkaIngressStartupPosition.fromLatest();
+
+  private OptionalConfig<String> consumerGroupId = OptionalConfig.withoutDefault();
+  private OptionalConfig<KafkaIngressDeserializer<T>> deserializer =
+      OptionalConfig.withoutDefault();
+  private OptionalConfig<String> kafkaAddress = OptionalConfig.withoutDefault();
+  private OptionalConfig<KafkaIngressAutoResetPosition> autoResetPosition =
+      OptionalConfig.withDefault(KafkaIngressAutoResetPosition.LATEST);
+  private OptionalConfig<KafkaIngressStartupPosition> startupPosition =
+      OptionalConfig.withDefault(KafkaIngressStartupPosition.fromLatest());
 
   private KafkaIngressBuilder(IngressIdentifier<T> id) {
     this.id = Objects.requireNonNull(id);
@@ -59,13 +63,13 @@ public final class KafkaIngressBuilder<T> {
 
   /** @param consumerGroupId the consumer group id to use. */
   public KafkaIngressBuilder<T> withConsumerGroupId(String consumerGroupId) {
-    this.consumerGroupId = Objects.requireNonNull(consumerGroupId);
+    this.consumerGroupId.set(consumerGroupId);
     return this;
   }
 
   /** @param kafkaAddress Comma separated addresses of the brokers. */
   public KafkaIngressBuilder<T> withKafkaAddress(String kafkaAddress) {
-    this.kafkaAddress = Objects.requireNonNull(kafkaAddress);
+    this.kafkaAddress.set(kafkaAddress);
     return this;
   }
 
@@ -100,7 +104,7 @@ public final class KafkaIngressBuilder<T> {
   public KafkaIngressBuilder<T> withDeserializer(
       Class<? extends KafkaIngressDeserializer<T>> deserializerClass) {
     Objects.requireNonNull(deserializerClass);
-    this.deserializer = instantiateDeserializer(deserializerClass);
+    this.deserializer.set(instantiateDeserializer(deserializerClass));
     return this;
   }
 
@@ -110,7 +114,7 @@ public final class KafkaIngressBuilder<T> {
    */
   public KafkaIngressBuilder<T> withAutoResetPosition(
       KafkaIngressAutoResetPosition autoResetPosition) {
-    this.autoResetPosition = Objects.requireNonNull(autoResetPosition);
+    this.autoResetPosition.set(autoResetPosition);
     return this;
   }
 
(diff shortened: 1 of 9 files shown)


tzulitai:
cc @igalshilman this is the follow-up to #5, concerning the config resolution issue.

statefun-kafka-io/src/main/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressBuilder.java
```diff
@@ -168,13 +174,20 @@ private Properties resolveKafkaProperties() {
     }
   }
 
+  private static <T> void overwritePropertyIfNamedConfigIsPresent(
```
igalshilman: Suggestion: Since `OptionalConfig` is a package private class, moving this method into the `OptionalConfig` might make it a bit more readable.

For example:
```
ConsumerConfig.GROUP_ID.upadte(properties, consumerGroupId);
```

Either way is good tho!

statefun-kafka-io/src/test/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressBuilderTest.java
```diff
@@ -0,0 +1,121 @@
...
+            .withKafkaAddress("localhost:8082")
+            .withTopic("topic")
+            .withConsumerGroupId("test-group")
+            .withDeserializer(NoOpDeserializer.class);
+
+    KafkaIngressSpec<String> spec = builder.build();
+
+    assertThat(spec.id(), is(DUMMY_ID));
+    assertThat(spec.type(), is(Constants.KAFKA_INGRESS_TYPE));
+    assertThat(spec.topics(), is(Collections.singletonList("topic")));
```
igalshilman: You can replace that with
```assertThat(spec.topics(), contains("topic"));```

File: statefun-kafka-io/src/test/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressBuilderTest.java [hunk 3]
```diff
@@ -0,0 +1,121 @@
...
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.junit.Test;
+
+public class KafkaIngressBuilderTest {
+
+  private static final IngressIdentifier<String> DUMMY_ID =
+      new IngressIdentifier<>(String.class, "ns", "name");
+
+  @Test
+  public void exampleUsage() {
```
igalshilman: I think that this test can be broken down a little bit, too number of separated tests.
one that validates that the topics are there, and one that valises that the properties are set currently, and one for the serializer.

statefun-kafka-io/src/test/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressBuilderTest.java
```diff
@@ -0,0 +1,121 @@
...
+            isMapOfSize(3),
+            hasProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:8082"),
+            hasProperty(ConsumerConfig.GROUP_ID_CONFIG, "test-group"),
+            hasProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "latest")));
+  }
+
+  @Test
+  public void namedMethodConfigValuesOverwriteProperties() {
+    Properties properties = new Properties();
+    properties.setProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, "should-be-overwritten");
```
igalshilman: I think that having 1 property is already enough.

statefun-kafka-io/src/test/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressBuilderTest.java
```diff
@@ -0,0 +1,121 @@
...
+        allOf(
+            hasProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:8082"),
+            hasProperty(ConsumerConfig.GROUP_ID_CONFIG, "test-group"),
+            hasProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest")));
+  }
+
+  @Test
+  public void defaultNamedMethodConfigValuesShouldNotOverwriteProperties() {
+    Properties properties = new Properties();
+    properties.setProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:8082");
```
igalshilman: I think that one property is already enough.

statefun-kafka-io/src/test/java/org/apache/flink/statefun/sdk/kafka/testutils/Matchers.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Objects;
+import java.util.Properties;
+import org.apache.flink.statefun.sdk.kafka.KafkaIngressStartupPosition;
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeMatcher;
+
+public class Matchers {
+
+  public static Matcher<KafkaIngressStartupPosition> startupPositionOf(
```
igalshilman: 👍 For using a custom matcher!
I'm wondering if it actually make senes to add `.equals()` to startup position, since they equality is well defined anyways.

statefun-kafka-io/src/main/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressSpec.java
```diff
@@ -75,12 +72,41 @@ public KafkaIngressStartupPosition startupPosition() {
     return startupPosition;
   }
 
-  private static void validateStartupPosition(
-      KafkaIngressStartupPosition startupPosition, @Nullable String groupIdConfig) {
-    if (startupPosition.isGroupOffsets() && groupIdConfig == null) {
+  private static Properties requireValidProperties(Properties properties) {
```
igalshilman: @tzulitai What are your thoughts on having the validation happen in the builder?

tzulitai: The only "downside" I had in mind when not implementing this in the builder was that it is slightly more dangerous implementing it there, because the timeliness of when the validation was done will affect the end result (before / after the properties resolution).
Just add one little bit of space to go wrong, although very trivial.
Being done in the `KafkaIngressSpec`, this should not be the case.

tzulitai: I don't have a strong stand on this though.
For a better separation of responsibility wise, it could make sense to still move this to the builder yes.

tzulitai: I don't have a strong stand on this though.
For a better separation of responsibility wise, it could make sense to still move this to the builder yes.

statefun-kafka-io/src/main/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressBuilder.java
tzulitai: Makes sense. Updated according to your suggestion.

statefun-kafka-io/src/test/java/org/apache/flink/statefun/sdk/kafka/testutils/Matchers.java
tzulitai: Good point, will change.

statefun-kafka-io/src/main/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressSpec.java
tzulitai: I'd like to keep this as is for now, but am not against changing it.
Please let me know if you'd like this changed after merging.

tzulitai:
Merging after Travis is green ...

PR #9, opened 2020-02-03
[FLINK-15851] [build] Add azure-pipelines.yml for Azure Pipelines CI setup

tzulitai:
This PR is based on #8, because it fixes the build for JDK 11. Only the second commit is relevant.

This is a first step in setting up a CI build pipeline for Stateful Functions. For an overview of all steps, please see description in the JIRA ticket.

This Azure Pipelines YAML configuration runs two jobs, one that builds with JDK 8, and the another with JDK 11.
Setting this up to also cover end-to-end tests is a follow-up task, tracked by umbrella ticket FLINK-15850.

---

Verifying that this works:
Please see an example triggered build summary / output logs at https://dev.azure.com/tzulitai/tzulitai.flink.statefun/_build/results?buildId=25

Note that this also verifies the JDK 11 build fix in #8.

PR #10, opened 2020-02-03
[FLINK-15851] [build] Add .travis.yml for Travis CI setup

tzulitai:
Alternative approach to #9 where Stateful Functions continue to use Travis for the time being.

This could make more sense considering:
- The Azure Pipelines setup with ASF is currently a bit problematic
- Travis is pretty sufficient for the current needs of the Stateful Functions project

---

Example build summary can be found here:
https://travis-ci.org/tzulitai/flink-statefun/builds/645335960

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,3 +27,5 @@ jobs:
       jdk: "openjdk11"
       script: mvn clean install
       name: build - jdk11
+
+# Test
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,13 @@ dist: xenial
 
 language: java
 
-jdk:
- - openjdk8
- - openjdk11
+jobs:
+  include:
+    - if: type in (pull_request, push)
+      jdk: "openjdk8"
+      script: mvn clean install
+      name: build - jdk8
+    - if: type in (pull_request, push)
+      jdk: "openjdk11"
+      script: mvn clean install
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,22 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+dist: xenial
+
+language: java
+
+jdk:
```
igalshilman: I'm not familiar with Travis+Github integration, is this is enough to trigger `mvn clean install` for commits pushed to a PR / master?
If so, then I'm +1 to merge this.

tzulitai: AFAIK it should be, but I guess it wouldn't hurt to define this more explicitly.
Will merge with an updated version that defines the mvn command explicitly!

tzulitai: Seems like the default command is `mvn test -B` only.
Will define jobs explicitly.

tzulitai:
Currently waiting on ASF Infra to enable Travis for the repository before merging this:
https://issues.apache.org/jira/browse/INFRA-19811

tzulitai:
Merging ...

PR #11, opened 2020-02-03
[hotfix][docs] Fix documentation build

sjwiesman:
This documentation is much stricter than Jekyll. I also removed the link to Marta's email. We should discuss adding the docs to a travis cron as a follow up.


tzulitai:
Thanks for fixing the documentation build @sjwiesman.
I've verified that it properly builds now locally.

I've opened a JIRA ticket to add documentation build to Travis cron per your suggestion:
https://issues.apache.org/jira/browse/FLINK-15874

tzulitai:
Merging ...

PR #12, opened 2020-02-04
[FLINK-15899] [build] Inherit from Apache Parent POM

tzulitai:
This is a first step towards making Stateful Functions release-ready.

See details about what settings the Apache Parent POM covers here: https://maven.apache.org/pom/asf/index.html.

As an end result of this change, we should be able to perform `mvn clean install -Papache-release` without issues.

PR #13, opened 2020-02-05
[FLINK-15910] [legal] Add NOTICE and LICENSE files

tzulitai:
The Apache Parent POM added in #12 already pulls in default LICENSE / NOTICE files for built Maven artifact jars.

This PR explicitly adds NOTICE / LICENSE files for the following distribution purposes:

**Canoncial source distribution**

0659c0d Adds NOTICE file to be shipped with the source distribution.
The only dependency we bundle in the source is copied & modified source code from the fastutils project (http://fastutil.di.unimi.it/), so this is mentioned.
See the copied code under `statefun-flink/statefun-flink-core`.

**Maven artifacts**

9161d70
The only artifact jar that bundles dependencies is `statefun-flink-distribution`, so an explicit NOTICE file is added for it under `META-INF`.
Additionally, since some BSD licensed dependencies are bundled, their respective licenses can also be found under `META-INF/licenses` to be bundled with the jar.


tzulitai:
cc @igalshilman

Also cc @zentol, if you find the time, would be great to have your input here to see if we're missing anything else legal wise.

statefun-flink/statefun-flink-distribution/src/main/resources/META-INF/NOTICE
```diff
@@ -0,0 +1,40 @@
+statefun-flink-distribution
+Copyright 2014-2020 The Apache Software Foundation
+
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
+
+This project bundles the following dependencies from (or other subprojects of) the owning top-level project, Apache Flink,
+under the Apache Software License 2.0 (http://www.apache.org/licenses/LICENSE-2.0.txt)
```
tzulitai: @zentol do you think we actually need this whole section regarding bundled `flink` / `flink-shaded` dependencies?

statefun-flink/statefun-flink-distribution/src/main/resources/META-INF/NOTICE
```diff
@@ -0,0 +1,40 @@
...
+
+- org.apache.flink:flink-annotations:1.10-SNAPSHOT
```
tzulitai: `1.10-SNAPSHOT` should be updated once Flink 1.10 is released and we no longer rely on the snapshots.

NOTICE
```diff
@@ -0,0 +1,9 @@
+Apache Flink
+Copyright 2014-2020 The Apache Software Foundation
+
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
+
+This project contains copied and modified source code that is licensed under the Apache Software License 2.0 (http://www.apache.org/licenses/LICENSE-2.0.txt) from the following projects:
+
+- fastutil (http://fastutil.di.unimi.it/) Copyright (c) - Sebastiano Vigna
```
zentol: Are the files in question tagged appropriately? For maintenance purposes you always want to be able to trace down the files from the NOTICE.

pom.xml
```diff
@@ -187,6 +187,8 @@ under the License.
                         <exclude>statefun-docs/docs/_templates/**</exclude>
                         <!-- Generated code  -->
                         <exclude>**/generated/**</exclude>
+                        <!-- Bundled license files -->
+                        <exclude>**/LICENSE*</exclude>
```
zentol: exclude source NOTICE as well?

statefun-flink/statefun-flink-distribution/src/main/resources/META-INF/NOTICE
zentol: If the shade-plugin is configured correctly then the Flink entries will be added automatically upon inclusion.

You want to add this transformer:
```
<transformer implementation="org.apache.maven.plugins.shade.resource.ApacheNoticeResourceTransformer">
	<projectName>Apache Flink-statefun</projectName>
	<encoding>UTF-8</encoding>
</transformer>
```

NOTICE
tzulitai: @zentol what kind of tagging do you have in mind?
Would a mention in NOTICE of where the copied files are in the source be sufficient?

pom.xml
tzulitai: @zentol It seems like NOTICE / LICENSE files without any suffixes are automatically ignored by the rat plugin already.

NOTICE
zentol: A small note in the javadocs + paths in the notice, to cover all our bases.

tzulitai: Got it, thanks

tzulitai:
Thanks for the review @zentol!
All comments have been addressed. I'll proceed to merge this once Travis is green.

PR #16, opened 2020-02-07
[FLINK-15945] Remove MULTIPLEX_FLINK_STATE config

tzulitai:
Prior to this PR, Stateful Functions allow enabling or disabling multiplex state, regardless of the state backend being used. This actually does not make sense, considering that:
* If the heap backend is used, there is no reason to multiplex state. Also taking into account that there are problems that exist for multiplexing state such as lack of support for state schema evolution.
* If RocksDB backend is used, it is highly encouraged to multiplex state, anyways.

Therefore, this PR removes the `MULTIPLEX_FLINK_STATE` configuration and simply piggybacks the behaviour of multiplexing state to the state backend being used, i.e. multiplex only when RocksDB state backend is used.

### DIFF
--- a/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/functions/Reductions.java
+++ b/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/functions/Reductions.java
@@ -151,6 +151,6 @@ final class Reductions {
     // TODO ideally, we should revisit how configuration is being passed to the
     // TODO operators to be available at runtime
     return backendClassName.equals(
-            "org.apache.flink.contrib.streaming.state.RocksDBKeyedStateBackend");
+        "org.apache.flink.contrib.streaming.state.RocksDBKeyedStateBackend");
   }
 }
--- a/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/functions/Reductions.java
+++ b/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/functions/Reductions.java
@@ -150,13 +150,7 @@ final class Reductions {
     // TODO to avoid additional dependencies on the Flink state backends
     // TODO ideally, we should revisit how configuration is being passed to the
     // TODO operators to be available at runtime
-    if (backendClassName.equals("org.apache.flink.runtime.state.heap.HeapKeyedStateBackend")) {
-      return false;
-    }
-    if (backendClassName.equals(
-        "org.apache.flink.contrib.streaming.state.RocksDBKeyedStateBackend")) {
-      return true;
-    }
-    throw new IllegalStateException("Unrecognized state backend type: " + backendClassName);
+    return backendClassName.equals(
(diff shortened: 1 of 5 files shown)


statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/functions/Reductions.java
```diff
@@ -145,4 +142,21 @@ void processEnvelopes() {
...
+
+    // TODO this is fragile and error-prone to classname changes, but we're doing this
+    // TODO to avoid additional dependencies on the Flink state backends
+    // TODO ideally, we should revisit how configuration is being passed to the
+    // TODO operators to be available at runtime
+    if (backendClassName.equals("org.apache.flink.runtime.state.heap.HeapKeyedStateBackend")) {
+      return false;
+    }
+    if (backendClassName.equals(
+        "org.apache.flink.contrib.streaming.state.RocksDBKeyedStateBackend")) {
```
igalshilman: @tzulitai I think that we only need this branch, 
and this method can be simplified to simply:
`return backendClassName.equals(
         "org.apache.flink.contrib.streaming.state.RocksDBKeyedStateBackend")`

I think that we shouldn't care about any other state backends.

In addition, can a state backend be NULL, or is there a default toy state backend that is used for the mini cluster or something a long these lines?

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/functions/Reductions.java
```diff
@@ -79,7 +76,7 @@ static Reductions create(
     container.add("keyed-state-backend", KeyedStateBackend.class, keyedStateBackend);
     container.add(new DynamicallyRegisteredTypes(statefulFunctionsUniverse.types()));
 
-    if (configuration.getBoolean(StatefulFunctionsJobConstants.MULTIPLEX_FLINK_STATE)) {
+    if (useMultiplexedState(keyedStateBackend)) {
```
igalshilman: 👍

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/functions/Reductions.java
igalshilman: Ok, I actually see that the build fails due to this :-)
I'd vote to only focus on RocksDB.

sjwiesman: @igalshilman, the statebackend cannot be null. The default is MemoryStateBackend which is the same implementation as FsStateBackend but checks are stored in JM memory.

tzulitai: @igalshilman 👍 Makes total sense. Will simply this branch.

^ the state backend is never `null` up to this point.

tzulitai:
I've updated to simplify the branches.
Will merge this once Travis agrees.

PR #22, opened 2020-02-11
[FLINK-16002][statefun][testutil] Add stateful functions test utilities

sjwiesman:
The stateful functions SDK does not have any dependencies on the Flink runtime. These test utilities provide a minimal synchronous runtime implementation to facilitate simple unit testing.

### DIFF
--- a/statefun-testutil/pom.xml
+++ b/statefun-testutil/pom.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+  http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>statefun-parent</artifactId>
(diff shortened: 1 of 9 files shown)


sjwiesman:
cc @igalshilman

statefun-testutil/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+            <groupId>org.apache.flink</groupId>
+            <artifactId>statefun-sdk</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-all</artifactId>
+            <version>1.3</version>
+        </dependency>
+        <dependency>
```
sjwiesman: Do you think we should actually bundle this or mark it as a compile dependency?

igalshilman: well, this entire artifact is "test dependency", so I think it needs to be in scope compile.

sjwiesman: done

igalshilman:
We had a discussion with Seth offline, and I definitely think there is need for this test kit.

To summarize the discussion: 
There are these two ways of approaching unit testing this (probably more)
a) would be just call the stateful function with some TestContext and make assertions on the side effects.
Something a long the lines of 
```
assertThat(functionUnderTest.invoke(context, message), allOf(
        hasOutgoingMessage( messageMatches(...) )
        hasOutgoingEgress( ... )
        hasDelayedMessages( ...)
); 
```
b) actually simulate some stuff happening, and later collecting the results
and asserting on the expected behaviour.

This PR leans towards (b) and I can definitely see the value of that after seeing some examples, so I'm +1 to bring this to shape and merge this, I also like the use of matchers they make the tests more readable!

sjwiesman:
Gordon, this package game out of a need to unit test the functions for my statefun workshop. If it helps, you can see the usage on on this branch for real world usage 

https://github.com/sjwiesman/flink-statefun/tree/workshop?files=1

statefun-testutil/src/main/java/org/apache/flink/statefun/testutils/function/TestContext.java
```diff
@@ -0,0 +1,198 @@
...
+    watermark += duration.toMillis();
+
+    while (!pendingMessage.isEmpty() && pendingMessage.peek().timer <= watermark) {
+      messages.add(pendingMessage.poll().envelope);
+    }
+
+    return processAllMessages();
+  }
+
+  private Map<Address, List<Object>> processAllMessages() {
```
tzulitai: mostly a nitpick:

I feel like the responsibility of this method should be separated from the `TestContext`, perhaps a `SynchronousReductions` class, which uses the `TestContext`.
Adding to message queue and draining it will reside in the `SynchronousReductions`. 

I think it's good as it it is though, so will still merge this as it. Just food for thought.

PR #25, opened 2020-02-17
[FLINK-16106] Add PersistedList state primitive

tzulitai:
This PR introduces a `PersistedList` state primitive to Stateful Functions.

- 3491fea: Adds the user-facing SDK classes
- 69e8ce6: Introduces the Flink-based accessor for `PersistedList`
- fc190ca: Completes the addition by letting the `StateBinder` support `PersistedList`

---

This can be verified by the new test `StateBinderTest#bindPersistedList()`.

### DIFF
--- a/statefun-sdk/src/main/java/org/apache/flink/statefun/sdk/state/PersistedAppendingBuffer.java
+++ b/statefun-sdk/src/main/java/org/apache/flink/statefun/sdk/state/PersistedAppendingBuffer.java
@@ -18,6 +18,7 @@
 package org.apache.flink.statefun.sdk.state;
 
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
 import javax.annotation.Nullable;
@@ -125,20 +126,18 @@ public final class PersistedAppendingBuffer<E> {
   }
 
   /**
-   * Gets the elements of the persisted buffer as an {@link Iterable}.
+   * Gets an unmodifiable view of the elements of the persisted buffer, as an {@link Iterable}.
    *
    * <p>This may return {@code null} if the buffer is empty or had been cleared (with {@link
    * #clear()}).
    *
-   * <p>Note that any modifications to the view does not modify the elements in the persisted
-   * buffer. To modify elements in the persisted buffer, you must use {@link #replaceWith(List)}.
-   *
-   * @return a modifiable view, as an {@link Iterable}, of the elements of the persisted buffer, or
-   *     {@code null} if the buffer is empty or had been cleared.
+   * @return an unmodifiable view, as an {@link Iterable}, of the elements of the persisted buffer,
+   *     or {@code null} if the buffer is empty or had been cleared.
    */
   @Nullable
   public Iterable<E> view() {
-    return accessor.view();
+    final Iterable<E> view = accessor.view();
+    return (view != null) ? new UnmodifiableViewIterable<>(view) : null;
   }
 
   /** Clears all elements in the persisted buffer. */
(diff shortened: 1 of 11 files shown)


tzulitai:
cc @igalshilman

tzulitai:
@igalshilman as discussed offline, I've changed the primitive name to `PersistedAppendingBuffer`, and also emphasized a bit in the Javadocs the contracts of all supported operations.
The most important part to review, IMO, would be the Javadocs of the operations (see method Javadocs of `PersistedAppendingBuffer`) so that we can pin the contracts that we want to be supporting.

statefun-sdk/src/main/java/org/apache/flink/statefun/sdk/state/PersistedAppendingBuffer.java
```diff
@@ -0,0 +1,189 @@
...
+  /**
+   * Gets the elements of the persisted buffer as an {@link Iterable}.
+   *
+   * <p>This may return {@code null} if the buffer is empty or had been cleared (with {@link
+   * #clear()}).
+   *
+   * <p>Note that any modifications to the view does not modify the elements in the persisted
+   * buffer. To modify elements in the persisted buffer, you must use {@link #replaceWith(List)}.
+   *
+   * @return a modifiable view, as an {@link Iterable}, of the elements of the persisted buffer, or
```
igalshilman: Is it modifiable or unmodifiable?

tzulitai: The view is modifiable, e.g. Iterator.remove() is supported (I got it wrong during our discussion yesterday). But the persisted state will not be modified.

Keep in mind that this is only true for RocksDB and not for heap - 
but my thinking is that if the contract is set such that the view is always a snapshot copy and `replace` needs to be always be used to modify the persisted state, then it would work for both backends.

igalshilman: Interesting. Does this mean that using RocksDB we can implement efficient head removal (poping the first element)
If this there case, then I think we should add a pop method, and change the view method to simply items().
If later we would decide to support heap backend, then we can provide a different accessory that supports pop() by calling
list.sublist(1, list.size())

igalshilman: Interesting. Does this mean that using RocksDB we can implement efficient head removal (poping the first element)
If this there case, then I think we should add a pop method, and change the view method to simply items().
If later we would decide to support heap backend, then we can provide a different accessory that supports pop() by calling
list.sublist(1, list.size())

tzulitai: no, RocksDB can not do efficient head removal.

When I say that the view is modifiable, essentially:
You can treat the view that you get as a copy (underneath its just an `ArrayList`). BUT this is only the case for RocksDB.

igalshilman: Got it, then in that case.
Let’s reinforce the contract by wrapping the iterate with a unmodifiable iterable.
Other then that, I’m good for merging this as is!

tzulitai: Taking one step back here:

The intention _is_ to have a modifiable copy, so that it'll be easier for users to remove elements from the view, and then call `replaceWith` afterwards with the modified view.

The comment about different between RocksDB and Heap, is that for heap since you don't really get a copy, when modifying the view you already modify the persisted state.

So, the only way to reinforce the contract is to actually always make a copy of the returned `Iterable` in Stateful Functions space, but that's a big performance hit and I'm not sure it's worth doing.

igalshilman: Having a modifiable view that support `.remove()`, that doesn't*  remove the actual element from the persisted storage is a confusing API in my mind, and can be easily avoided by wrapping the returned Iterable with an `UnmodifieableIterable` and clearly document this that this is a read only view.

To add to the confusion 🙈  `.remove()` can remove the elements when using the heap state backend (potentially so, depends on rather or not a snapshot is happening concurrently etc')

PR #28, opened 2020-02-20
[FLINK-16159] [tests, build] Add verification integration test + integrate with Maven build

tzulitai:
This PR achieves the following goals:
- Adds a `statefun-integration-tests` module that is meant to maintain integration tests in the future.
- Adds `statefun-sanity-itcase` to the integration tests module, which includes a simple application used for sanity verification and a JUnit-driven test that performs the verification over a docker setup.
- Adds a Maven build profile so that the integration tests can be run with a simple command: `mvn clean verify -Prun-integration-tests`

---

### The verification app and test

The verification app (`SanityVerificationModule`) is a simple application used for sanity verification.
- The application reads commands from a Kafka ingress
- Has multiple functions binded (currently 2 in this PR) that reacts to the commands (see class-level Javadoc) of `SanityVerificationModule` for a full description on the set of commands)
- Reflects any state updates in the functions back to a Kafka egress.

The Junit-driven test does the following:
- Uses Testcontainers (https://www.testcontainers.org/) to start a Kafka broker, builds the image for the verification app on the fly and also starts a master + worker container for the app.
- Writes some commands to the Kafka ingress topic
- Reads state outputs for the Kafka egress topic, and verifies that they are correct

Right now the test scenario does not have any randomization to it.
We may consider to add that in the future as a follow-up.

---

### Maven `run-integration-tests` build profile and structure setup

With this PR, to add new container-based integration tests in the future, a developer has to do the following:
- Add a new sub-module under `statefun-integration-tests`
- Add source code for the Stateful Functions application to test against
- Add a test class named with the pattern `*ITCase` that setups the containers using testcontainers, and implements the test verification logic as a JUnit test. The classname pattern is important because only then would it be skipped by the unit test phase, and only kicks in with the `run-integration-tests` profile.

To build while also running the integration tests, one simply does:
`mvn clean verify -Prun-integration-tests` in the project root directory.

This always re-builds the base Stateful Functions image before running the ITCases.

---

### Integrating with CI

This is not yet integrated with CI.
We want to do that after we have more confidence that the tests are reliable.
Once we decide to do that, it would be as simple as enabling the build profile in the build commands.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,6 +57,7 @@ under the License.
         <module>statefun-quickstart</module>
         <module>statefun-docs</module>
         <module>statefun-testutil</module>
+        <module>statefun-end-to-end-tests</module>
     </modules>
 
     <properties>
--- a/pom.xml
+++ b/pom.xml
@@ -57,6 +57,12 @@ under the License.
         <module>statefun-quickstart</module>
         <module>statefun-docs</module>
         <module>statefun-testutil</module>
+
+        <!--
+            Always build the end-to-end tests module last,
+            since the module builds the base Stateful Functions image with the
+            built project artifacts before running the end-to-end tests.
+        -->
         <module>statefun-end-to-end-tests</module>
     </modules>
 
--- a/statefun-end-to-end-tests/pom.xml
+++ b/statefun-end-to-end-tests/pom.xml
@@ -31,4 +31,49 @@ under the License.
     <modules>
         <module>statefun-sanity-itcase</module>
     </modules>
+
+    <profiles>
+        <profile>
+            <!--
+                Build profile for running end-to-end tests in this module.
+                This first builds the base Stateful Functions image before running the ITCases.
+                Docker must be installed for this profile to work.
+            -->
+            <id>run-end-to-end-tests</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.codehaus.mojo</groupId>
+                        <artifactId>exec-maven-plugin</artifactId>
+                        <version>1.6.0</version>
+                        <executions>
+                            <execution>
+                                <phase>pre-integration-test</phase>
+                                <goals>
+                                    <goal>exec</goal>
+                                </goals>
+                                <inherited>false</inherited>
+                            </execution>
+                        </executions>
+                        <configuration>
+                            <executable>${user.dir}/tools/docker/build-stateful-functions.sh</executable>
+                        </configuration>
+                    </plugin>
+                    <plugin>
+                        <artifactId>maven-failsafe-plugin</artifactId>
+                        <version>2.22.0</version>
+                        <executions>
+                            <execution>
+                                <goals>
+                                    <goal>integration-test</goal>
+                                    <goal>verify</goal>
+                                </goals>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
+
 </project>
--- a/statefun-end-to-end-tests/pom.xml
+++ b/statefun-end-to-end-tests/pom.xml
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+  http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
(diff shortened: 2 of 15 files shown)


statefun-integration-tests/statefun-sanity-itcase/src/test/java/org/apache/flink/statefun/itcases/sanity/SanityVerificationITCase.java
```diff
@@ -0,0 +1,287 @@
...
+          "Type index is out of bounds. Max index: " + (Constants.FUNCTION_TYPES.length - 1));
+    }
+    return FnAddress.newBuilder().setType(typeIndex).setId(fnId).build();
+  }
+
+  // =================================================================================
+  //  Kafka ingress / egress serde
+  // =================================================================================
+
+  public static final class FnAddressSerializerDeserializer
```
igalshilman: Is it possible to replace all of these with a single `ProtobufSerializer<T extends Message>`,
not a big deal with it is not easy doable.

statefun-integration-tests/statefun-sanity-itcase/src/test/java/org/apache/flink/statefun/itcases/sanity/SanityVerificationITCase.java
```diff
@@ -0,0 +1,287 @@
...
+import org.testcontainers.images.builder.ImageFromDockerfile;
+
+/**
+ * Sanity verification integration test based on the {@link SanityVerificationModule} application.
+ *
+ * <p>The integration test setups Kafka brokers and the verification application using Docker, sends
+ * a few commands to Kafka to be consumed by the application, and finally verifies that outputs sent
+ * to Kafka from the application are correct.
+ */
+public class SanityVerificationITCase {
```
igalshilman: With what parallelism does this job starts?
Do you think should the sanity job start with parallelism > 1 ?

pom.xml
```diff
@@ -55,6 +55,7 @@ under the License.
         <module>statefun-examples</module>
         <module>statefun-flink</module>
         <module>statefun-quickstart</module>
+        <module>statefun-integration-tests</module>
```
igalshilman: It seems like and end to end test, don't you think?

statefun-integration-tests/statefun-sanity-itcase/src/test/java/org/apache/flink/statefun/itcases/sanity/SanityVerificationITCase.java
```diff
@@ -0,0 +1,287 @@
...
+    final String kafkaAddress = kafka.getBootstrapServers();
+    final ExecutorService kafkaIoExecutor = Executors.newCachedThreadPool();
+
+    kafkaIoExecutor.submit(new ProduceCommands(kafkaAddress));
+    Future<List<StateSnapshot>> stateSnapshotOutputs =
+        kafkaIoExecutor.submit(new ConsumeStateSnapshots(kafkaAddress));
+
+    assertThat(
+        stateSnapshotOutputs.get(1, TimeUnit.MINUTES),
+        hasItems(
```
igalshilman: Suggestion:
It could be nicer if we had thecommands present at the same level that the verification is present.
For example:

```
assertThat(
       sending(
           modifiyAction(address1, 100),
           modifiyAction(address2, 200))
     , results(
               stateSnapshot(address1, 100)
               stateSnapshot(address2, 200))
        );
```

It is fine the way it is right now, I'm sure will iterate on these kind of tests many times in the future

tzulitai:
@igalshilman FYI I'll address the rest of the comments next week.
Thanks for the review so far, will ping you again once it is ready.

statefun-integration-tests/statefun-sanity-itcase/src/test/java/org/apache/flink/statefun/itcases/sanity/SanityVerificationITCase.java
tzulitai: Turns out there is a nice way to do that!
See fdcdef1c1a4697bbecb0d9949d002dbe76551f44 for the fix.

tzulitai:
@igalshilman all your comments have been addressed, and the PR is ready for review. Since the follow-up commit history got a bit out of control, I figured that it'll be easier if I rebased everything and squashed to understandable commits, and therefore had to force push.

Besides some of the more minor comments:
- I've introduced our own `StatefulFunctionsAppContainers` test rule that is based on Testcontainers. This would remove a lot of the cumbersome setup code from the actual test code, and makes it easier to setup multiple workers in the end-to-end tests. See 
86ccc87. I can see this as a common utility to be used by other end-to-end tests in the future.

- The test now runs in parallelism 2. This is done by adding a `flink-conf.yaml` with `parallelism.default=2` to the built application image, and having 2 workers in the `StatefulFunctionsAppContainers` test rule.

- Concerning a more readable assertion of Kafka inputs and outputs (your comment [here](https://github.com/apache/flink-statefun/pull/28#discussion_r381997568)), I gave that a try and came up with a `KafkaIOVerifier` utility. See bb2e0cc. It's not 100% good though; needs a good way to verify end-of-output situations (please see the `TODO` in `KafkaIOVerifier`).

igalshilman:
Thanks for the update @tzulitai, this looks really good!
Let's merge that and slowly build confidence by adding e2e test cases to cover existing functionality.

tzulitai:
Thanks a lot, merging ...

PR #32, opened 2020-02-24
[FLINK-16149][hotfix][core] Use GlobalConfiguration.loadConfiguration() on StreamPlanEnvironment

sjwiesman:
This is a follow up to #27 

Because the statefun-flink-launcher is based on JobClusterEntrypoint it
pulls the StreamGraph using a StreamPlanEnvironment. In this case, the
StreamExecutionEnviornment configuration is set to an empty object.

The only way to pull the actual flink-conf in this case is via
GlobalConfiguration.loadConfiguration(). This is safe because
StreamPlanEnvironment implies job based deployment. In all other cases
the configuration is still pulled via reflection.

I've tested this using statefun images and deploying against a local session cluster setting different configurations and verifying that they were used at runtime. 

cc @tzulitai @igalshilman

### DIFF
--- a/statefun-end-to-end-tests/statefun-sanity-itcase/pom.xml
+++ b/statefun-end-to-end-tests/statefun-sanity-itcase/pom.xml
@@ -29,6 +29,7 @@ under the License.
 
     <properties>
         <testcontainers.version>1.12.5</testcontainers.version>
+        <flink.version>1.10.0</flink.version>
     </properties>
 
     <dependencies>
@@ -82,6 +83,14 @@ under the License.
             <groupId>org.testcontainers</groupId>
             <artifactId>kafka</artifactId>
             <version>${testcontainers.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <!-- Flink Config -->
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-core</artifactId>
+            <version>${flink.version}</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/statefun-end-to-end-tests/statefun-sanity-itcase/src/test/java/org/apache/flink/statefun/itcases/sanity/SanityVerificationITCase.java
+++ b/statefun-end-to-end-tests/statefun-sanity-itcase/src/test/java/org/apache/flink/statefun/itcases/sanity/SanityVerificationITCase.java
@@ -27,6 +27,7 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
+import org.apache.flink.configuration.Configuration;
 import org.apache.flink.statefun.itcases.sanity.generated.VerificationMessages.Command;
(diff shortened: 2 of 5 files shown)


statefun-end-to-end-tests/statefun-sanity-itcase/src/test/java/org/apache/flink/statefun/itcases/sanity/SanityVerificationITCase.java
```diff
@@ -61,14 +62,20 @@
...
+
   @Rule
   public KafkaContainer kafka =
       new KafkaContainer(CONFLUENT_PLATFORM_VERSION)
           .withNetworkAliases(Constants.KAFKA_BROKER_HOST);
 
   @Rule
   public StatefulFunctionsAppContainers verificationApp =
-      new StatefulFunctionsAppContainers("sanity-verification", 2)
+      new StatefulFunctionsAppContainers("sanity-verification", 2, flinkConf)
```
tzulitai: It'll be nicer if this is supplied like:
```

File: statefun-end-to-end-tests/statefun-sanity-itcase/pom.xml [hunk 2]
```diff
@@ -29,6 +29,7 @@ under the License.
 
     <properties>
         <testcontainers.version>1.12.5</testcontainers.version>
+        <flink.version>1.10.0</flink.version>
```
tzulitai: I think we don't need to define this here. Parent POM already defines `flink.version` so we should inherit that.

statefun-end-to-end-tests/statefun-sanity-itcase/src/test/java/org/apache/flink/statefun/itcases/sanity/SanityVerificationITCase.java
tzulitai: [suggested as a follow-up, no need for this PR]

It'll be nicer if this is supplied like:
```
public StatefulFunctionsAppContainers verificationApp =
    new StatefulFunctionsAppContainers("sanity-verification", 2)
        .withModuleGlobalConfiguration("kafka-broker", kafka.getBootstrapServers())
        .withConfiguration(ConfigOption option, configValue)
```

And by default the `StatefulFunctionsAppContainers` just only has the configs in the base template `flink-conf.yaml`.

That would eliminate the need for explicitly setting a network alias for the `KafkaContainer` (i.e. less boilerplate code in these e2e tests).

This would require lazy construction of the containers on `beforeTest`, a slightly bigger change that is out-of-scope here. So I suggest not including it in this PR.

statefun-end-to-end-tests/statefun-sanity-itcase/pom.xml
tzulitai: Ah scratch that 😅 It's only defined in `statefun-flink`.
nevermind then :)

PR #34, opened 2020-02-26
[FLINK-16272] Move e2e test utilities to a statefun-e2e-tests-common module

tzulitai:
This PR achieves 2 simple things:
- Move `StatefulFunctionsAppContainers` / `KafkaIOVerifier` / `ProtobufSerDe` to a common `statefun-e2e-tests-common` module so that new e2e tests can use them.
- Rename all e2e related classes / packages / modules to `*e2e*`. End-to-end test modules living under `statefun-e2e-tests` should now be named `statefun-*-e2e`, and test classes should be named `*E2E.java`.

---

### Verifying

Run `mvn clean verify -Prun-e2e-tests`

PR #35, opened 2020-02-26
[FLINK-16274] Support inlined configuration in StatefulFunctionsAppContainers

tzulitai:
This PR is based on #34. The first 4 commits are not relevant.

---

### Goal

This PR achieves the following things:
- Change the `StatefulFunctionsAppContainers` to a more "builder-ish" approach, which the user provides settings on how the containers are built, and only on `before()` the test we build the containers and application images. This would allow a more fluent API to inline configuration settings.
- Add inline configuration setting methods, like so:
```
@Rule
public StatefulFunctionsAppContainers verificationApp =
    new StatefulFunctionsAppContainers("app-name", numWorkers)
        .withModuleGlobalConfiguration("key", "value");
        .withConfiguration(ConfigOption<T> option, T value);
```
This also eliminates the need for users of `StatefulFunctionsAppContainers` to maintain a `flink-conf.yaml` in their classpath, as they can now build the configuration fully programmatically with the new inline configuration methods.

---

### Verifying

This is only changes in tests.
Run `mvn clean verify -Prun-e2e-tests` to verify changes.


sjwiesman:
LGTM +1

tzulitai:
Thanks for the review @sjwiesman!
After some offline discussions with @igalshilman, we also came to the conclusion that this would be a useful addition. Please see the discussions in the JIRA ticket.

Proceeding to merge this ...

PR #36, opened 2020-02-26
[FLINK-16123] Add auto-routable Protobuf Kafka ingress

tzulitai:
This PR adds support for a Protobuf Kafka ingress that automatically routes values as Protobuf `Any` messages to target functions using the Kafka record's key (as UTF8 strings) as the function id.

The YAML definition looks like the following:
```
ingress:
  meta:
    type: org.apache.flink.statefun.sdk.kafka/routable-protobuf-kafka-connector
    id: com.mycomp.foo/bar
  spec:
    address: kafka-broker:9092
    consumerGroupId: my-group-id
    topics:
      - topic: topic-1
        typeUrl: com.googleapis/com.mycomp.foo.MessageA
        targets:
          - com.mycomp.foo/function-1
          - com.mycomp.foo/function-2
      - topic: topic-2
        typeUrl: com.googleapis/com.mycomp.foo.MessageB
        targets:
          - com.mycomp.foo/function-2
    autoOffsetResetPosition: earliest
    startupPosition:
      type: earliest
    properties:
      - foo.config: bar
```

Summary:
- For each topic, users define the Protobuf type URL of the message type in that topic, and the list of target functions to route records of that topic to.
- Users must use as key the target function ids, as a UTF8 string.

When binding this ingress to a module, `JsonModule` automatically binds a `AutoRoutableProtobufRouter` for the ingress. Therefore, users do not need to explicitly define a router for the ingress in the YAML.

---

### Verifying

- A unit test `RoutableProtobufKafkaSourceProviderTest` is added to demonstrate example YAML and usage.
- An E2E test for the feature will follow-up after this PR.

### DIFF
--- a/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/jsonmodule/JsonModule.java
+++ b/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/jsonmodule/JsonModule.java
@@ -50,7 +50,9 @@ import org.apache.flink.statefun.flink.core.httpfn.HttpFunctionProvider;
 import org.apache.flink.statefun.flink.core.httpfn.HttpFunctionSpec;
 import org.apache.flink.statefun.flink.core.jsonmodule.FunctionSpec.Kind;
 import org.apache.flink.statefun.flink.core.jsonmodule.Pointers.Functions;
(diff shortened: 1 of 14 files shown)


tzulitai:
Thanks a lot for the review @igalshilman!
I'll open a follow-up JIRA for the suggestion you made around the `RoutingConfig` - lets address that as a follow-up.

All comments have been addressed, proceeding to merge this!

PR #42, opened 2020-03-02
[FLINK-16390][sdk] Add view and clear methods to PersistedTable

sjwiesman:
Adds common collection methods to PersistedTable. These methods are already available on PersistedAppendingBuffer. 

Note: I have `view` return an empty iterable instead of null. This is different than the default return value for `MapState#entries` but is consistent with `PersistedAppendingBuffer#view`.

### DIFF
--- a/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/state/FlinkTableAccessor.java
+++ b/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/state/FlinkTableAccessor.java
@@ -17,6 +17,7 @@
  */
 package org.apache.flink.statefun.flink.core.state;
 
+import java.util.Map;
 import java.util.Objects;
 import org.apache.flink.api.common.state.MapState;
 import org.apache.flink.statefun.sdk.state.TableAccessor;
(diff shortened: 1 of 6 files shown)


statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/state/FlinkTableAccessor.java
```diff
@@ -55,4 +57,19 @@ public void remove(K key) {
       throw new RuntimeException(e);
     }
   }
+
+  @Override
+  public Iterable<Map.Entry<K, V>> view() {
+    try {
+      Iterable<Map.Entry<K, V>> entries = handle.entries();
+      return entries == null ? Collections.emptyList() : entries;
```
tzulitai: Actually, I think Flink user `MapState` never returns `null` for `entries()` / `keys()` / `values()`.

The actual `MapState` handle that the user gets is always a `UserFacingMapState` in Flink that wraps the actual RocksDB / heap map state handle, which uses an empty map state if the wrapped state handle returns `null` for those methods.

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/state/FlinkTableAccessor.java
```diff
@@ -55,4 +57,19 @@ public void remove(K key) {
       throw new RuntimeException(e);
     }
   }
+
+  @Override
+  public Iterable<Map.Entry<K, V>> view() {
```
tzulitai: Do you intend the returned iterable to be modifiable (i.e. the iterator supports `remove()`)?
If so, maybe we should just call this `entries()` instead.

We called the method "view" in the `PersistedAppendingBuffer` to emphasize the fact that the returned iterable is unmodifiable.

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/state/FlinkTableAccessor.java
sjwiesman: Looks like I have some Flink jobs I can remove some null checks from :)

sjwiesman:
@tzulitai I've updated based on your comments. I originally went with view to be consistent with `PersistedAppendingBuffer`, but since they have different semantics I switched it back to entries.

tzulitai:
Thanks for the update @sjwiesman.

+1 LGTM, merging ..

PR #44, opened 2020-03-03
[FLINK-16395] [legal] Add OkHttp to NOTICE of statefun-flink-distribution

tzulitai:
cc @igalshilman


tzulitai:
Thanks! Merging ...

PR #45, opened 2020-03-03
[FLINK-16396] [kafka] Add Protobuf message for YAML-ized Kafka egress

tzulitai:
This adds a Protobuf message to be used by the YAML-ized Kafka egress (WIP).
I'm eagerly opening this PR already because there's a dependency in the Python SDK that would require this definition in place.


igalshilman:
The protobuf message looks good to me.
+1

tzulitai:
Thanks! Merging ...

PR #47, opened 2020-03-03
[FLINK-16326] [core] Eagerly validate strictly required Flink configurations

tzulitai:
This PR achieves the following:
- 7850755 .Make it clearer in the base `flink-conf.yaml` template what configurations should not be touched. Also add `execution.checkpointing.max-concurrent-checkpoints: 1` to the template since Stateful Functions currently does not support concurrent checkpoints.
- 5706462 Add eager validation of the strictly required Flink configs to `StatefulFunctionsConfig` class.
- d8c2739 Adapt the Flink Harness to these changes.

---

### Verifying

I've verified this for both Flink Harness executions as well as end-to-end containerized executions.
- For Flink Harness executions, I executed the harness example (which fails without d8c2739).
- For end-to-end executions, I did `mvn clean verify -Prun-e2e-tests`

### DIFF
--- a/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/StatefulFunctionsConfig.java
+++ b/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/StatefulFunctionsConfig.java
@@ -25,16 +25,22 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import org.apache.flink.configuration.ConfigOption;
 import org.apache.flink.configuration.ConfigOptions;
 import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.CoreOptions;
 import org.apache.flink.configuration.GlobalConfiguration;
 import org.apache.flink.configuration.MemorySize;
 import org.apache.flink.configuration.description.Description;
+import org.apache.flink.statefun.flink.core.exceptions.StatefulFunctionsInvalidConfigException;
 import org.apache.flink.statefun.flink.core.message.MessageFactoryType;
 import org.apache.flink.statefun.sdk.spi.StatefulFunctionModule;
+import org.apache.flink.streaming.api.environment.ExecutionCheckpointingOptions;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.apache.flink.streaming.api.environment.StreamPlanEnvironment;
 import org.apache.flink.util.InstantiationUtil;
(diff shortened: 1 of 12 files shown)


statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/StatefulFunctionsConfig.java
```diff
@@ -243,4 +246,34 @@ public void addAllGlobalConfigurations(Map<String, String> globalConfigurations)
   public void setGlobalConfiguration(String key, String value) {
     this.globalConfigurations.put(key, value);
   }
+
+  private static void validateStrictlyRequiredConfigs(Configuration configuration) {
```
igalshilman: Can you extract the validation logic to a separate class?
It can be an inner static class for now, or a separate class with validation specific logic.

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/StatefulFunctionsConfig.java
```diff
@@ -243,4 +246,34 @@ public void addAllGlobalConfigurations(Map<String, String> globalConfigurations)
...
+    final Set<String> parentFirstClassloaderPatterns =
+        parentFirstClassloaderPatterns(configuration);
+    if (!parentFirstClassloaderPatterns.contains("org.apache.flink.statefun")
+        && !parentFirstClassloaderPatterns.contains("org.apache.kafka")
```
igalshilman: shouldn't that be an `||`instead of `&&` ?

tzulitai: Yes you're right, it should be `||` :)

sjwiesman: nit: I think the semantics would be clearer if you put the strings in a static final List and then did `parentFirstClassloaderPatterns.containsAll`.

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/StatefulFunctionsConfig.java
```diff
@@ -243,4 +246,34 @@ public void addAllGlobalConfigurations(Map<String, String> globalConfigurations)
...
+          CoreOptions.ALWAYS_PARENT_FIRST_LOADER_PATTERNS_ADDITIONAL,
+          "Must contain org.apache.flink.statefun, org.apache.kafka, and com.google.protobuf");
+    }
+
+    final int maxConcurrentCheckpoints =
+        configuration.get(ExecutionCheckpointingOptions.MAX_CONCURRENT_CHECKPOINTS);
+    if (maxConcurrentCheckpoints != 1) {
+      throw new StatefulFunctionsInvalidConfigException(
+          ExecutionCheckpointingOptions.MAX_CONCURRENT_CHECKPOINTS,
+          "Value must be 1, since Stateful Functions currently does not support concurrent checkpoints.");
```
sjwiesman: nit
```suggestion
          "Value must be 1, Stateful Functions does not support concurrent checkpoints.");
```

tzulitai:
Thanks for the reviews @igalshilman @sjwiesman!
All your comments have been addressed with the suggestions.

Will proceed to merge this now ..

PR #49, opened 2020-03-04
[FLINK-16396] [kafka] Allow null keys for the YAML generic Kafka egress

tzulitai:
A followup-fix for #46, which allows the key to be `null`.
In this case, Kafka will perform round-robin partitioning (assuming the user did not configure a partitioner in the Kafka client properties) on the written producer records.


tzulitai:
cc @igalshilman

tzulitai:
Thanks, merging ...

PR #50, opened 2020-03-04
[FLINK-16415] [core] Add format versioning to YAML-ized modules

tzulitai:
This PR adds a required `version` field to YAML-ized modules.

The starting format version will be `1.0`, following the conventions of systems like docker-compose.

The changelog commits should be self-explanatory on how this is implemented.
Note: From the looks of the code, the `Pointers` and `JsonModule` classes could use some refactoring, but we can revisit that once we introduce more format versions in the future.

---

### Verifying

The `JsonModuleTest` has been adapted to cover this change.
Same for the E2E test - the `RoutableKafkaE2E` uses a YAML module, so that is covered there as well.


tzulitai:
cc @igalshilman

igalshilman:
Looks good to me!

tzulitai:
Thanks, merging ...

PR #51, opened 2020-03-07
Make PersistedTable maintain the insertion order.

danp11:


igalshilman:
Hi @danp11, thank you for your continued interest in statefun and your contribution!

However as per `Flink`'s [contribution guide ](https://flink.apache.org/contributing/contribute-code.html) any contribution starts with a JIRA issue submitted, having an issue submitted allows contributors / committers to provide an early advice and potentially save some development effort :-)

Unfortunately this feature can not be efficiently implemented for the distributed case (you've only added support to the non fault tolerant accessor which is used for tests)
The accessor that is used at runtime, is leveraging Flink state that is provided by a state backend,
and there, there is no guarantee about the insertion order.

danp11:
All good and Thanks for your detalied response.

/Dan

Den sön 8 mars 2020 21:07Igal Shilman <notifications@github.com> skrev:

tzulitai:
I agree with @igalshilman.
I'm also not following the motivation behind having insertion order in `PersistedTable`.
Is it because you want to use it as a queue?
If so, I think a dedicated persisted state primitive would be more suitable for that. Such a thing would definitely require some upfront communication, preferably in JIRAs, before opening pull requests.

tzulitai:
@danp11 I'll close this PR for now. Let me know if you disagree with this, otherwise as suggested above, lets take a step back and move the discussions to the JIRA.

danp11:
Hi again, yes my thought was to use it as a queue for publishing of events.

However, it might be a better option to block the publisher until the
publishing is finished and then there is no need for a queue in this
usecase.

Sorry for beeing a bit eager here😀

/Dan



Den mån 9 mars 2020 09:36Tzu-Li Tai <notifications@github.com> skrev:

PR #55, opened 2020-03-10
 [FLINK-16518] [kafka] Set client properties as strings in KafkaSinkProvider

tzulitai:
The put method is strongly discourage to be used on `Properties` as a bad practice, since it allows putting non-string values.

This has already caused a bug, where a long was put into the properties, while Kafka was expecting an integer:
```
org.apache.kafka.common.config.ConfigException: Invalid value 100000 for configuration transaction.timeout.ms: Expected value to be a 32-bit integer, but it was a java.lang.Long
	at org.apache.kafka.common.config.ConfigDef.parseType(ConfigDef.java:669)
	at org.apache.kafka.common.config.ConfigDef.parseValue(ConfigDef.java:471)
	at org.apache.kafka.common.config.ConfigDef.parse(ConfigDef.java:464)
	at org.apache.kafka.common.config.AbstractConfig.<init>(AbstractConfig.java:62)
	at org.apache.kafka.common.config.AbstractConfig.<init>(AbstractConfig.java:75)
	at org.apache.kafka.clients.producer.ProducerConfig.<init>(ProducerConfig.java:396)
	at org.apache.kafka.clients.producer.KafkaProducer.<init>(KafkaProducer.java:326)
	at org.apache.kafka.clients.producer.KafkaProducer.<init>(KafkaProducer.java:298)
	at org.apache.flink.streaming.connectors.kafka.internal.FlinkKafkaInternalProducer.<init>(FlinkKafkaInternalProducer.java:76)
```


tzulitai:
cc @igalshilman

igalshilman:
Thanks @tzulitai,
Looks good to me.

PR #56, opened 2020-03-12
[FLINK-16558] [doc] Reword project website tagline

tzulitai:
This PR changes the project tagline in the docs to:

image

It previously read: "A framework for stateful distributed applications by the original creators of Apache Flink®." which seems a bit outdated now, since the project is maintained by the community as a whole.

### DIFF
--- a/statefun-docs/docs/_templates/layout.html
+++ b/statefun-docs/docs/_templates/layout.html
@@ -8,6 +8,6 @@
 
 	{{ super() }}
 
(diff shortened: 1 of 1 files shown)


tzulitai:
After some offline discussion with @morsapaes and @igalshilman,
we decided to only change the last bit to be `powered by Apache Flink` for now.
The tagline itself deserves a separate proper discussion of its own.

merging ...

PR #58, opened 2020-03-14
[FLINK-16254] Support -p / --parallelism command option

tzulitai:
This PR adds support for specifying parallelism for Stateful Function applications using the `-p / --parallelism` command.

The new behaviour as as follows:
- If `-p` is defined in the command line, then that value is always used as the parallelism.
- Otherwise, the `parallelism.default` value in `flink-conf.yaml` is used, which by default is 1 if not present.

---

### Changelog

- 27e7abc Adds parsing logic of the parallelism option from the command line, and adds the property to `StatefulFunctionsClusterConfiguration`.
- 07c8dd2 Resolves the parallelism to use in `StatefulFunctionsJobGraphRetriever`, and uses the resolved value to create the JobGraph.
- ee61047 Adapt the E2E tests, so that they use the `-p` command to specify parallelism.

---

### Verfying

The changes to the E2E tests in ee61047 verifies this change: `mvn clean verify -Prun-e2e-tests`

### DIFF
--- a/statefun-e2e-tests/statefun-e2e-tests-common/src/main/java/org/apache/flink/statefun/e2e/common/StatefulFunctionsAppContainers.java
+++ b/statefun-e2e-tests/statefun-e2e-tests-common/src/main/java/org/apache/flink/statefun/e2e/common/StatefulFunctionsAppContainers.java
@@ -31,7 +31,6 @@ import java.util.Objects;
 import javax.annotation.Nullable;
 import org.apache.flink.configuration.ConfigOption;
 import org.apache.flink.configuration.Configuration;
-import org.apache.flink.configuration.CoreOptions;
 import org.apache.flink.configuration.GlobalConfiguration;
 import org.apache.flink.statefun.flink.core.StatefulFunctionsConfig;
 import org.junit.rules.ExternalResource;
@@ -142,10 +141,6 @@ public final class StatefulFunctionsAppContainers extends ExternalResource {
     this.network = Network.newNetwork();
     this.appName = appName;
     this.numWorkers = numWorkers;
-
-    if (numWorkers > 1) {
-      dynamicProperties.set(CoreOptions.DEFAULT_PARALLELISM, numWorkers);
-    }
   }
 
   public StatefulFunctionsAppContainers dependsOn(GenericContainer<?> container) {
@@ -180,7 +175,7 @@ public final class StatefulFunctionsAppContainers extends ExternalResource {
   protected void before() throws Throwable {
     final ImageFromDockerfile appImage =
         appImage(appName, dynamicProperties, classpathBuildContextFiles);
-    this.master = masterContainer(appImage, network, dependentContainers, masterLogger);
+    this.master = masterContainer(appImage, network, dependentContainers, numWorkers, masterLogger);
     this.workers = workerContainers(appImage, numWorkers, network);
 
     master.start();
@@ -261,13 +256,15 @@ public final class StatefulFunctionsAppContainers extends ExternalResource {
       ImageFromDockerfile appImage,
       Network network,
       List<GenericContainer<?>> dependents,
+      int numWorkers,
       @Nullable Logger masterLogger) {
     final GenericContainer<?> master =
         new GenericContainer(appImage)
             .withNetwork(network)
             .withNetworkAliases(MASTER_HOST)
             .withEnv("ROLE", "master")
-            .withEnv("MASTER_HOST", MASTER_HOST);
+            .withEnv("MASTER_HOST", MASTER_HOST)
+            .withCommand("-p " + numWorkers);
 
     for (GenericContainer<?> dependent : dependents) {
       master.dependsOn(dependent);
--- a/statefun-flink/statefun-flink-launcher/src/main/java/org/apache/flink/statefun/flink/launcher/StatefulFunctionsJobGraphRetriever.java
+++ b/statefun-flink/statefun-flink-launcher/src/main/java/org/apache/flink/statefun/flink/launcher/StatefulFunctionsJobGraphRetriever.java
@@ -25,6 +25,7 @@ import java.net.URI;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.List;
+import org.apache.flink.api.common.ExecutionConfig;
 import org.apache.flink.api.common.JobID;
 import org.apache.flink.client.program.PackagedProgram;
 import org.apache.flink.client.program.PackagedProgramUtils;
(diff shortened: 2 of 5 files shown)


statefun-flink/statefun-flink-launcher/src/main/java/org/apache/flink/statefun/flink/launcher/StatefulFunctionsJobGraphRetriever.java
```diff
@@ -113,4 +124,10 @@ private PackagedProgram createPackagedProgram() {
       throw new RuntimeException("Unable to construct a packaged program", e);
     }
   }
+
+  private static int resolveParallelism(int parallelism, Configuration configuration) {
```
igalshilman: This seems a bit tricky,
Consider the case where a user set a value 3, in Flink-conf.yaml
But for testing purposes they provide -p with a value equal to PATALLEISM_DEFAULT.

With what degree of parallelism would the job start?

statefun-e2e-tests/statefun-e2e-tests-common/src/main/java/org/apache/flink/statefun/e2e/common/StatefulFunctionsAppContainers.java
```diff
@@ -142,10 +141,6 @@ public StatefulFunctionsAppContainers(String appName, int numWorkers) {
     this.network = Network.newNetwork();
     this.appName = appName;
     this.numWorkers = numWorkers;
-
-    if (numWorkers > 1) {
-      dynamicProperties.set(CoreOptions.DEFAULT_PARALLELISM, numWorkers);
```
igalshilman: Do we actually know that the test succeed because of the -p
This seems to also set the option in the configurations?

statefun-flink/statefun-flink-launcher/src/main/java/org/apache/flink/statefun/flink/launcher/StatefulFunctionsJobGraphRetriever.java
tzulitai: The parallelism used will be the value provided via `-p`.

Note that `ExecutionConfig.PARALLELISM_DEFAULT` is `-1`. It's merely a flag to indicate that no parallelism was set.

tzulitai: But I think you have a point -
We should probably when parsing the command line, make sure that the value provided via `-p` is `>= 1`. That would avoid the problem you mentioned.

statefun-e2e-tests/statefun-e2e-tests-common/src/main/java/org/apache/flink/statefun/e2e/common/StatefulFunctionsAppContainers.java
tzulitai: I verified that the `-p` option is respected, due to the test logs:
```
14:24:17,713 INFO  org.apache.flink.statefun.e2e.sanity.SanityVerificationE2E    - 2020-03-14 06:24:17,745 INFO  org.apache.flink.runtime.executiongraph.ExecutionGraph        - Source: org.apache.flink.e2e.sanity-commands-ingress -> router (commands) (1/2) (629795504f9adeabfcdf7bd54a19b208) switched from DEPLOYING to RUNNING.
14:24:17,737 INFO  org.apache.flink.statefun.e2e.sanity.SanityVerificationE2E    - 2020-03-14 06:24:17,769 INFO  org.apache.flink.runtime.executiongraph.ExecutionGraph        - Source: org.apache.flink.e2e.sanity-commands-ingress -> router (commands) (2/2) (39160ca2105d2a831cb436043ed42667) switched from DEPLOYING to RUNNING.
14:24:17,953 INFO  org.apache.flink.statefun.e2e.sanity.SanityVerificationE2E    - 2020-03-14 06:24:17,985 INFO  org.apache.flink.runtime.executiongraph.ExecutionGraph        - feedback (1/2) (2ac2575233c0b11d009dec544c6dd8e3) switched from DEPLOYING to RUNNING.
14:24:18,070 INFO  org.apache.flink.statefun.e2e.sanity.SanityVerificationE2E    - 2020-03-14 06:24:18,103 INFO  org.apache.flink.runtime.executiongraph.ExecutionGraph        - feedback (2/2) (44b80a1071b85a4ea7b2647196199276) switched from DEPLOYING to RUNNING.
14:24:18,071 INFO  org.apache.flink.statefun.e2e.sanity.SanityVerificationE2E    - 2020-03-14 06:24:18,104 INFO  org.apache.flink.runtime.executiongraph.ExecutionGraph        - feedback-union -> functions -> Sink: org.apache.flink.e2e.sanity-state-snapshots-egress (2/2) (4103da85baa83f85ab864cf6756fb801) switched from DEPLOYING to RUNNING.
14:24:18,109 INFO  org.apache.flink.statefun.e2e.sanity.SanityVerificationE2E    - 2020-03-14 06:24:18,142 INFO  org.apache.flink.runtime.executiongraph.ExecutionGraph        - feedback-union -> functions -> Sink: org.apache.flink.e2e.sanity-state-snapshots-egress (1/2) (97c8e328a747f0cd071ee468eb4a9a3b) switched from DEPLOYING to RUNNING.
```

As you can see, each operator has 2 subtasks being run.

igalshilman: What I mean is, how do you know that this is thanks to the -p and not thanks to the line highlighted above?

statefun-flink/statefun-flink-launcher/src/main/java/org/apache/flink/statefun/flink/launcher/StatefulFunctionsJobGraphRetriever.java
igalshilman: Aha! In that case -1 works!
Then consider this comment to be resolved.

statefun-e2e-tests/statefun-e2e-tests-common/src/main/java/org/apache/flink/statefun/e2e/common/StatefulFunctionsAppContainers.java
igalshilman: Ok, scratch that, I’ve looked at the wrong commit.
These lines are actually removed.

tzulitai:
Thanks for the review Igal!
Merging ...

PR #59, opened 2020-03-16
[FLINK-16603] Fix quickstart Maven archetype

tzulitai:
The quickstart Maven archetype was not working due to a few issues:

- the archetype `pom.xml` not being included for filtering, causing the `@project.version@` string not being replaced with the project version.
- the archetype `pom.xml` did not define `modelVersion`

---

### Verifying

An integration-test phase test has been added to verify that projects generated with the archetype can be compiled.


tzulitai:
Thanks for reviewing Igal!
I've fixed the problem you mentioned - `src/main/resources` wasn't being filtered previously, now it should work.

PR #69, opened 2020-03-24
[FLINK-16738] [py] Add kinesis_egress_record builder method to Python SDK

tzulitai:
This PR adds a `kinesis_egress_record` method to the Python SDK, so that we hide the business of building a `KinesisEgressRecord` message from the user.
It is similar to how we added the `kafka_egress_record` method to the Python SDK.


igalshilman:
LGTM.

tzulitai:
Thanks, merging ...

PR #78, opened 2020-03-28
[FLINK-16838] Change base image name and apply versioning

tzulitai:
This PR has the following end-goal in mind:
- Examples / E2E tests / quickstart archetype built or run from snapshot versions (i.e. from `master` or `release-2.0` branches) should always run against a locally-built StateFun base image, built from the source of said snapshot version.
- Released examples / E2E tests / quickstart archetypes, should not require users to locally build a StateFun image, assuming that we will have official images published to Docker Hub.

This PR does a few things to accomplish that:

1. Let the image build script `tools/docker/build-distribution.sh` build images tagged with the current source version. i.e.,
`2.1-SNAPSHOT` when built from the snapshot `master` branch, `2.0-SNAPSHOT` when built from the snapshot `release-2.0` branch, or
simply `2.0.0` if built from a officially released source distribution

2. Update Dockerfiles of all examples / E2E tests / quickstart archetype to use a specific version tag. This has the effect that, if those were built from a released distribution, the image may be pulled directly from Docker Hub (when we publish the images after the release). Otherwise, if those were built from a snapshot version, they would use a locally-built snapshot image.

On the side, this PR also uses the opportunity to rename the image name from `statefun` to `flink-statefun`, with the following reasoning:
- It would be more consistent with the naming convention of other distributions, like the Python SDK
- Lets the image name to show association with Flink

---

## Verifying

I verified the changes by:
- Running all the examples that used images built from StateFun base image, and all ran without problems
- Run all end-to-end tests, using `mvn clean verify -Prun-e2e-tests`
- Created new project from quickstart archetype - the generated skeleton project has the correct Dockerfile:
```
FROM flink-statefun:2.1-SNAPSHOT

RUN mkdir -p /opt/statefun/modules/my-app
COPY target/my-app*jar /opt/statefun/modules/my-app/
```

### DIFF
--- a/statefun-e2e-tests/statefun-routable-kafka-e2e/src/test/resources/Dockerfile
+++ b/statefun-e2e-tests/statefun-routable-kafka-e2e/src/test/resources/Dockerfile
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM statefun
+FROM flink-statefun:2.1-SNAPSHOT
 
 RUN mkdir -p /opt/statefun/modules/statefun-routable-kafka-e2e
 COPY statefun-routable-kafka-e2e*.jar /opt/statefun/modules/statefun-routable-kafka-e2e/
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ submitted to a Flink cluster.
 Below is an example Dockerfile for building an image for an application called `statefun-example`:
 
 ```
-FROM statefun
+FROM flink-statefun[:version-tag]
 
 RUN mkdir -p /opt/statefun/modules/statefun-example
 COPY target/statefun-example*jar /opt/statefun/modules/statefun-example/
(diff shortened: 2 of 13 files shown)


statefun-e2e-tests/statefun-routable-kafka-e2e/src/test/resources/Dockerfile
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM statefun
+FROM flink-statefun:2.1-SNAPSHOT
```
tzulitai: Note:
these will be changed to `FROM flink-statefun:2.0-SNAPSHOT` when backporting these changes to the snapshot `release-2.0` branch.

On release candidates / releases, these will be e.g. `FROM flink-statefun:2.0.0`

tzulitai: The updated `update_branch_version.sh` script takes care of updating these strings automatically.

igalshilman:
Thanks!
This looks good to me.

tzulitai:
Thanks for the fast review Igal.

Merging ...

PR #79, opened 2020-03-28
[FLINK-16730][docs] Add walkthrough distribution as build step

sjwiesman:
The python walkthrough is based on the python greeter example and includes skeleton code to quickly get started. Python does not have an equivalent to maven archetype so I wanted to provide a zip download.

We add a custom build step on `mvn package` for the `statefun-python-greeter` module that zips the package and copyies the zip file into `docs/download`. The zip is excluded from source releases.  

cc @tzulitai


statefun-examples/statefun-python-greeter/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+specific language governing permissions and limitations
+under the License.
+-->
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <artifactId>statefun-parent</artifactId>
```
tzulitai: this should be `statefun-examples`

tools/releasing/update_branch_version.sh
```diff
@@ -77,6 +77,9 @@ perl -pi -e "s#version_title: \"$OLD_VERSION\"#version_title: \"$NEW_VERSION\"#"
 find . -name 'Dockerfile*' -type f -exec perl -pi -e "s#FROM flink-statefun:$OLD_VERSION#FROM flink-statefun:$NEW_VERSION#" {} \;
 perl -pi -e "s#VERSION_TAG=$OLD_VERSION#VERSION_TAG=$NEW_VERSION#" tools/docker/build-stateful-functions.sh
 
+# Rebuild python walkthrough dist with updated versions
+./../docs/create_python_walkthrough.sh
```
tzulitai: This should be the relative path from project root:
```suggestion
tools/docs/create_python_walkthrough.sh
```

tzulitai:
After some time thinking about this, I think we probably shouldn’t rush adding the zip dist building as a Maven build step.

The problem is: every time you do a `mvn clean package`, it generates a git diff with the regenerated zip file, even if the contents are actually identical. It’ll be a bit confusing.

We could add that build step behind a Maven profile that developers have to enable to build the doc walkthrough dists, but that’s almost the same work for devs as just manually executing the script.

For now, lets proceed like so:
- Only generate the wallthrough zip in the docs buildbot and have that published with the docs. That’s the only time it’ll really be needed to be checked for an update, anyways. When building the docs locally, that download link will be broken, but I think that’s fine for now.
- That means I’ll be removing the Maven build step + zip from the repo + remain excluded in the .gitignore.
- Lets revisit this once we realize we need more and more Python zip dists in the docs

Sorry for the back and forth on the solution, but after some thinking I think this is the safest approach to take first.

sjwiesman:
Sounds good, thanks

PR #83, opened 2020-03-31
[FLINK-16879] [build] Disable the source-release-assembly execution goal

tzulitai:
We should be using the `apache-release` build profile defined in the Apache Parent POM, with one exclusion: do not run the `source-release-assembly` execution goal.

We have our own scripts to create the source release, which correctly excludes binaries from distribution tarball. That script can be found under `tools/releasing/create_source_release.sh`.

---

## Verifying

Build the project using `mvn clean install -Papache-release`.
Under `project-root/target/`, you should not find any source release packages, only POM files.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -85,6 +85,41 @@ under the License.
         </dependency>
     </dependencies>
 
+    <profiles>
+        <!--
+            We're reusing the apache-release build profile defined in the Apache Parent POM,
+            with one exclusion: do not run the source-release-assembly execution goal.
+            We have our own scripts to create the source release, which correctly excludes
+            binaries from distribution tarball.
+            The script can be found under tools/releasing/create_source_release.sh.
+        -->
+        <profile>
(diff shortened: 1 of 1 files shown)


tzulitai:
cc @rmetzger 
This fixes the problem you found with RC4 of 2.0.0.

This does not require a cancellation of the current ongoing RC, because the artifacts in the new Maven repository for RC4 is still built with the same commit hash.

rmetzger:
Thanks a lot for proposing this fix.
I guess we don't have this problem in apache/flink because we are not using the "apache-release" profile for building the releases there?

tzulitai:
@rmetzger correct, Flink defines its own `release` profile, but as far as I can tell, it uses essentially the same plugins used as the `apache-release` profile, minus the `source-release-assembly` execution.

tzulitai:
Thanks for the review, merging this ...

PR #88, opened 2020-04-01
Licensing final fixes / touches for 2.0.0

tzulitai:
None of the fixes here are hard blockers for the release.

This PR contains the following fixes for licenses:
- 0ecc467 excludes the bundled NOTICE / LICENSE pulled in from Kafka. The contents in those are misleading, since it cites a binary dependency on jersey, but this is neither reflected in the dependency graph, nor are any jersey files bundled. The approach here follows how `flink-sql--kafka-connector` treats this specific case.
- b4927f1 and 34991df makes sure that we do not have any duplicate license files under the name `LICENSE.txt` / `license.txt` pulled in. This is ok, since all non-ASLv2 licenses are manually included under the `META-INF/licenses` directory.
- e18030f This is actually a NOTICE file merging issue that should be fixed upstream in the Flink Kinesis connector. We're applying the fix here before it is fixed upstream.

These should be backported to `release-2.0`.


statefun-examples/statefun-ridesharing-example/statefun-ridesharing-example-simulator/pom.xml
```diff
@@ -169,6 +178,24 @@ under the License.
...
                         </filter>
+                        <filter>
+                            <artifact>org.apache.kafka:*</artifact>
+                            <excludes>
+                                <exclude>kafka/kafka-version.properties</exclude>
+                                <exclude>LICENSE</exclude>
+                                <!-- Does not contain anything relevant.
+                                    Cites a binary dependency on jersey, but this is neither reflected in the
+                                    dependency graph, nor are any jersey files bundled. -->
+                                <exclude>NOTICE</exclude>
```
rmetzger, line 189: Are we including the contents of Kafka's `NOTICE` file somewhere?

We have to forward the Kafka NOTICE file, even if we disagree with it's contents (at least that's how I understand the ASL 2.0)

tzulitai: To address that we can manually merge it to our NOTICE file (its not getting picked up by the `ApacheNoticeResourceTransformer`), but as far as I can tell Flink itself is also excluding the contents completely.

tzulitai: How it is being done in Flink may be wrong though -
I won't be against merging the contents to our NOTICE file, to be safe.

statefun-examples/statefun-ridesharing-example/statefun-ridesharing-example-simulator/src/main/resources/META-INF/licenses/LICENSE.javax.annotation-api
```diff
@@ -360,400 +360,3 @@ litigation relating to this License shall be subject to the jurisdiction
 of the Federal Courts of the Northern District of California and the
 state courts of the State of California, with venue lying in Santa Clara
 County, California.
-
```
rmetzger, line 363: I'm not an expert, but here's my opinion:

I don't feel comfortable editing a license we are forwarding:
The CDDL states 
```
Except as provided in Section 4.3, no one other than the license steward has the right to modify this License.
```
I'm not sure if "this License" means only the CDDL, or the LICENSE file of javax annotation api.

Why can't we forward the javax annotation api LICENSE file (unmodified, as requested by the license), but state in our NOTICE file, that we are using the javax annotation api under the terms of the CDDL?

I believe this approach is less risky (remember `Oracle is the initial license steward`)

statefun-examples/statefun-ridesharing-example/statefun-ridesharing-example-simulator/pom.xml
StephanEwen: We have to only forward parts of the NOTICE relevant to the parts we redistribute, according so ASLv2.

rmetzger: Okay, thanks.
Then this review item is resolved.

statefun-examples/statefun-ridesharing-example/statefun-ridesharing-example-simulator/src/main/resources/META-INF/licenses/LICENSE.javax.annotation-api
StephanEwen: Hmmm, interesting point. My point came from other dual-licensed projects (like Jackson was in the past) that included two license files (ASLv2, LGPL) and you forwarded the one you selected. I was thinking to mirror that here, but I don't know if the assumption here is that you have to forward the information about being dual licensed.

tzulitai: I think I agree with @rmetzger here.
Modying the LICENSE seems risky.
Relying on our NOTICE to clearly state which license we're using the dependency under the terms of, should be sufficient.

StephanEwen:
I think Robert have a valid point to not remove parts of the `javax.annotation` license (sorry for pointing in the wrong direction earlier).

Other than that, +1 to these changes.

tzulitai:
Thanks a lot for the prompt reviews, merging this to `master` and `release-2.0`.

PR #89, opened 2020-04-01
[FLINK-16925] remove unused CSE files

alpinegizmo:
docs/cse.xml and docs/annotations.xml aren't actually used, and their presence here might confuse someone.


StephanEwen:
Looks good, thanks!

Merging this...

PR #93, opened 2020-04-02
[FLINK-16948][docs] Change docs to better highlight embedded vs remote module configurations

sjwiesman:
Update the docs based on user feedback to better highlight how to use io modules in embedded vs remote modules.


sjwiesman:
@knaufk Would you want to take a look at this.

knaufk:
LGTM

PR #99, opened 2020-04-16
[FLINK-17193] [k8s-python-example] build-example.sh improvements

aroch:
- The helm template was printed out instead of being redirected to the resource YAML file.

- In case the SDK distribution was not previously built - instead of failing the build, build the distribution and continue.

- Abort script on failure. Previously, on failure the script would continue to run all steps and finish in an erroneous state. Also this saves 'if' statements to check exit codes after commands execute.

### DIFF
--- a/statefun-examples/statefun-python-k8s-example/build-example.sh
+++ b/statefun-examples/statefun-python-k8s-example/build-example.sh
@@ -1,4 +1,7 @@
 #!/bin/bash
+
+set -e
+
 #
 # Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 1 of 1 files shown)


sjwiesman:
Hi @aroch, thank you for fixing this. I will go ahead and merge it.

In the future, could you please open a Jira ticket for this type of change. It helps us keep track on of what is being worked on in the project.

sjwiesman:
Thank you!

aroch:
Thanks @sjwiesman! Just opened an issue and updated accordingly. Was not completely sure about filling the JIRA. Would appreciate if you can take a look.

sjwiesman:
Merged.

PR #100, opened 2020-04-19
[FLINK-17060][docs] Replace "{flink}" with the Apache Flink trademark

authuir:
Replace "{flink}" with the Apache Flink trademark


tzulitai:
LGTM! Thanks for the fix @authuir.
Merging this ...

PR #101, opened 2020-04-21
[FLINK-17287][github] Disable merge commit button

zentol:
(I know it is already disabled, this is just for consistency and to introduce the asf.yaml concept to all repos)


tzulitai:
LGTM, merging! Thanks @zentol.

tzulitai:
LGTM, merging! Thanks @zentol.

PR #106, opened 2020-05-07
[FLINK-16985] Stateful Functions Examples have their specific job names

abc863377:
Stateful Functions Examples add their specific job names.

Greeter-example
Python-greeter-example
Python-kubernetes-example
Ridesharing-example
Shopping-cart-example

### DIFF
--- a/tools/docker/flink-distribution-template/conf/flink-conf.yaml
+++ b/tools/docker/flink-distribution-template/conf/flink-conf.yaml
@@ -31,3 +31,5 @@ state.backend.rocksdb.timer-service.factory: ROCKSDB
 state.checkpoints.dir: file:///checkpoint-dir
 state.backend.incremental: true
 taskmanager.memory.process.size: 4g
+
+statefun.flink-job-name: Statefun Application


abc863377:
@tzulitai  Can you please review it?

igalshilman:
Hi there,
Thanks a lot for opening your first PR for this project, and welcome :-)

I think that everything we will put in the examples folks would accidentally think that this is mandatory and copy to their projects.
I’d really like the Dockerfile to stay as simple as possible at this moment, to lower the entrance barrier for StateFun.

Did you find the lack of a specific job name confusing for you?

tzulitai:
I originally opened FLINK-16985.
I wouldn't say lacking the specific job name is confusing, but maybe rounds up a better user experience for new users who are just coming to this repo for the first time, running the application, and then looking at the Flink Web UI to see what's happening behind the scenes.

But I agree that adding this extra bit to the example Dockerfiles can be a bit distracting from a "introductory examples" point of view.

igalshilman:
Hi @abc863377, I think that there is a consensus, proceed with the change here, but I'd like to suggest the following: can you set the `statefun.flink-job-name` key  at `tools/docker/flink-distribution-template/conf/flink-conf.yaml` so that people who would like to change the job name, could know what to modify.

The suggest change is to add:
```
statefun.flink-job-name: Statefun Application
```

abc863377:
Hi @igalshilman , 
Thanks for the suggestion. I'll follow your advice and keep Dockerfile simple.

abc863377:
Hi, @igalshilman ,
Sorry. I am a Git beginner.

Now I removed the old commit. and updated the new statefun.flink-job-name configuration to the distribution template.

Can you please review it?

Thanks a lot.

igalshilman:
Hi @abc863377,
No worries, thanks a lot for your first contribution to this project!

sjwiesman:
Merging . . .

PR #110, opened 2020-05-14
[FLINK-17611] Unix domain socket connection between worker and function

slinkydeveloper:
Hi all,
This pr implements https://issues.apache.org/jira/browse/FLINK-17611 .

To enable a unix domain socket connection in a remote module, use `uds` field:

```
      - function:
          meta:
            kind: http
            type: com.bar/world
          spec:
            endpoint: /statefun
            uds: /hello/world.sock
            states:
              - seen_count
            maxNumBatchRequests: 10000
```

`uds` field defines the file name, endpoint the path

I'm wondering, I test only the module parsing, is there any way to add an integration test here? If yes, should i do that in this pr or in a follow up?


sjwiesman:
Thank you for working on this feature! I haven't taken a look at the code, Igal or Gordon are more qualified to look at this part of the code base, but I wanted to provide some early feedback. 

1) I am fairly certain that flinks shaded netty, which is available from flink-core, contains a unix socket implementation. Does it make sense to use that instead of pulling in a new dependency? 

2) If we do add a new dependency to statefun-flink-core we need to update the NOTICE file accordingly[1]. Can you please check how junixsocket-core is licensed? 

3) You can take a look at flink-e2e-tests module to see how to write end to end tests. They are dockerized and the project provides some utilities to simplify things. Do note that #108 is refactoring the utility classes so it might make sense to wait for that to be merged in first. 

[1] https://github.com/apache/flink-statefun/blob/master/statefun-flink/statefun-flink-core/src/main/resources/META-INF/NOTICE

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/httpfn/HttpFunctionProvider.java
```diff
@@ -41,12 +45,32 @@ public RequestReplyFunction functionOfType(FunctionType type) {
...
+        spec.states(), spec.maxNumBatchRequests(), buildHttpClient(spec));
+  }
+
+  private RequestReplyClient buildHttpClient(HttpFunctionSpec spec) {
+    // We need to build a UDS HTTP client
+    if (spec.unixDomainSocket() != null) {
+      OkHttpClient specificClient =
+          sharedClient
+              .newBuilder()
+              .socketFactory(new AFUNIXSocketFactory.FactoryArg(spec.unixDomainSocket()))
```
tzulitai, line 58: Is there a reason that you chose not to implement a custom UDS socket factory like in the example at https://github.com/square/okhttp/tree/master/samples/unixdomainsockets/src/main/java/okhttp3/unixdomainsockets?

I'm asking only because we should try to have the `statefun-flink-core` dependencies as slim as possible, as it directly influences the size of our base distributions. It might also not be too big of a deal if the `junixsocket-core` library is small.

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/httpfn/HttpFunctionProvider.java
```diff
@@ -41,12 +45,32 @@ public RequestReplyFunction functionOfType(FunctionType type) {
...
+              // Enable HTTP/2 if available (uses H2 upgrade),
+              // otherwise fallback to HTTP/1.1
+              .protocols(Collections.singletonList(Protocol.HTTP_2))
+              .callTimeout(spec.maxRequestDuration())
+              .build();
+
+      return new HttpRequestReplyClient(
+          // Only the path matters!
+          HttpUrl.get(URI.create(spec.endpoint().getPath())), specificClient);
```
tzulitai, line 67: For easier readability: maybe it makese sense to add to the spec class a getter method which just returns the path string without the scheme.

slinkydeveloper:
> As @sjwiesman mentioned, using the flink-e2e-tests module, you can write a full-blown end-to-end test, where you verify by writing to ingresses / asserting outputs in egresses. However, I'm not sure for this feature we actually need such a heavy test. Perhaps a more scoped integration test, where you only specifically test the integration between a UDS-enabled HttpRequestReplyClient against a mock web server (e.g. the MockWebServer provided by OkHttp looks promising) would be sufficient. This is also something I'm trying to define more clearly in the project: what makes sense to write an end-to-end test with the flink-e2e-tests utility, and for what a smaller scoped integration test is more reasonable.

That's a good idea, but can i find somewhere such test for standard over TCP http connection? Maybe i can just take that one and parametrize it

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/httpfn/HttpFunctionProvider.java
slinkydeveloper: I didn't used the uds library proposed by okhttp `jnr-unixsocket` because it doesn't look in a very good shape (last release is 0.31 https://github.com/jnr/jnr-unixsocket/releases)

I'll try to find out if we can use only `junixsocket-core`

slinkydeveloper: Ok so this factory comes from `common` module, that is the least module we can use because it's the one that contains the `java.net.Socket` implementation

slinkydeveloper:
@sjwiesman junixsocket is apache 2 https://github.com/kohlschutter/junixsocket/blob/master/LICENSE what should i write on the NOTICE?

tzulitai:
> @sjwiesman junixsocket is apache 2 https://github.com/kohlschutter/junixsocket/blob/master/LICENSE
> 
> what should i write on the NOTICE? I didn't "copied and modified source code" like the it.unimi package

The NOTICE file you should append to is this one:
https://github.com/apache/flink-statefun/blob/master/statefun-flink/statefun-flink-distribution/src/main/resources/META-INF/NOTICE

Dependencies of `statefun-flink-core` will be bundled within the fat distribution jar, so we need acknowledge that.
Good that it is ASL 2.0, then all you need to do is list is along with the other listed ASL 2.0 deps.

igalshilman:
@sjwiesman switching to netty would be much more involved change, and perhaps we would replace okhttp with netty down the road, but for now I'm okay with adding this dependency.

igalshilman:
I wouldn't go with an e2e test, it would be an overkill.
I think that having a mini integration test (independent of any stateful function logic) 
that connects to a mock web server via uds (as @tzulitai suggested) and basically replies a simple string should be enough.
following the lines of [this example](https://github.com/square/okhttp/blob/master/samples/unixdomainsockets/src/main/java/okhttp3/unixdomainsockets/ClientAndServer.java)

slinkydeveloper:
@igalshilman if you're looking for an async io transport, i suggest to go with https://vertx.io/docs/vertx-core/java/ . It's a little layer on top of netty that vastly simplifies the usage of the HTTP, providing additional features OOTB like client connection pooling and so on.

slinkydeveloper:
> One thing I would recommend doing before merging is to manually verify the change by modifying statefun-python-greeter-example to connect over the unix domain socket.

TBH I've tried to build that example but for some reason i have build failures with python sdk... i probably messed up something with my py env

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/httpfn/HttpFunctionProvider.java
```diff
@@ -41,12 +45,32 @@ public RequestReplyFunction functionOfType(FunctionType type) {
...
+  private RequestReplyClient buildHttpClient(HttpFunctionSpec spec) {
+    // We need to build a UDS HTTP client
+    if (spec.unixDomainSocket() != null) {
+      OkHttpClient specificClient =
+          sharedClient
+              .newBuilder()
+              .socketFactory(new AFUNIXSocketFactory.FactoryArg(spec.unixDomainSocket()))
+              // Enable HTTP/2 if available (uses H2 upgrade),
+              // otherwise fallback to HTTP/1.1
+              .protocols(Collections.singletonList(Protocol.HTTP_2))
```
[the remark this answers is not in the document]
slinkydeveloper: Should i do it in a separate PR or this one is fine?

statefun-flink/statefun-flink-core/src/test/java/org/apache/flink/statefun/flink/core/jsonmodule/JsonModuleTest.java
```diff
@@ -95,6 +96,7 @@ public void testEgresses() {
 
   private static StatefulFunctionModule fromPath(String path) {
     URL moduleUrl = JsonModuleTest.class.getClassLoader().getResource(path);
+    assertThat(moduleUrl, not(nullValue()));
```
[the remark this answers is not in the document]
slinkydeveloper: Separate commit or separate PR?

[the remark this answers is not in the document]
slinkydeveloper: Don't we squash the PR when we merge it? otherwise i rebase it a bit to look cleaner :smile:

igalshilman:
@slinkydeveloper You can use the already published Python SDK in PyPI (`pip install apache-flink-statefun`) it should be able to handle requests from your branch, since nothing has been changed.
The only thing missing is to configure the http server to bind to the uds. I think`gunicorn  --bind unix://...` should work.

statefun-flink/statefun-flink-core/src/test/java/org/apache/flink/statefun/flink/core/jsonmodule/JsonModuleTest.java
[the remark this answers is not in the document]
igalshilman: That’s fine, don’t worry about it.
We don’t necessarily squash, a single PR can produce multiple commits and they will be merged as is.
That’s why we try to separate the changes, i.e: did the thing, added test, refactored the test etc’
But don’t worry about it now, this is for your next PR ;)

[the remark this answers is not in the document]
slinkydeveloper: Ok thanks! I'll prefer to just squash this one

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/httpfn/HttpFunctionProvider.java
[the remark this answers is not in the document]
igalshilman: This one is fine!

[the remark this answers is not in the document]
slinkydeveloper: Done

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/httpfn/OkHttpUtils.java
```diff
@@ -38,6 +41,9 @@ static OkHttpClient newClient() {
         .followRedirects(true)
         .followSslRedirects(true)
         .retryOnConnectionFailure(true)
+        // Enable HTTP/2 if available (uses H2 upgrade),
+        // otherwise fallback to HTTP/1.1
+        .protocols(Collections.singletonList(Protocol.HTTP_2))
```
igalshilman, line 46: It seems like okhttp has HTTP_2  already in the default protocol list, so this is not needed. 
Moreover the above code throws an exception (when specifying HTTP_2 you must also specify HTTP_1) but if you would look at the default builder you would see that okhttp already has HTTP_2 and HTTP_1 in the list of default protocols.

As an attempt to "make it work" I've artificially added the prefix `http://localhost/` to the endpoint, what do you think about that (it is technically "localhost")?

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/httpfn/HttpFunctionProvider.java
```diff
@@ -41,12 +48,56 @@ public RequestReplyFunction functionOfType(FunctionType type) {
...
+          sharedClient
+              .newBuilder()
+              .socketFactory(
+                  new AFUNIXSocketFactory.FactoryArg(splittedFilePathAndEndpoint.getKey()))
+              .callTimeout(spec.maxRequestDuration())
+              .build();
+
+      return new HttpRequestReplyClient(
+          // Only the path matters!
+          HttpUrl.get(URI.create(splittedFilePathAndEndpoint.getValue())), specificClient);
```
igalshilman, line 73: There are two unfortunate things going on here:
1) `HttpUrl.get` would silently return `null` if it was unable to parse the URI
2) the endpoint only (rightfully so) contains the path without the schema and `HttpUrl` doesn't consider it to be valid.

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/httpfn/HttpFunctionProvider.java
```diff
@@ -41,12 +48,56 @@ public RequestReplyFunction functionOfType(FunctionType type) {
...
+
+      // We need to split the path in order to get the sock file and the path after the sock file
+      Map.Entry<String, String> splittedFilePathAndEndpoint =
+          splitFilePathAndEndpointForUDS(spec.endpoint());
+
+      OkHttpClient specificClient =
+          sharedClient
+              .newBuilder()
+              .socketFactory(
+                  new AFUNIXSocketFactory.FactoryArg(splittedFilePathAndEndpoint.getKey()))
```
igalshilman, line 67: Unfortunately this does not seems to be working.
I get the following stack trace:
It seems like okhttp is trying to set the socket timeout, before connecting (that makes senes) but the library 
expects first a connection and then setting a timeout (which also makes senes)

```
orker_1           | Caused by: java.net.SocketException: Not open
worker_1           | 	at org.newsclub.net.unix.AFUNIXSocketImpl.validFdOrException(AFUNIXSocketImpl.java:424)
worker_1           | 	at org.newsclub.net.unix.AFUNIXSocketImpl.setOption(AFUNIXSocketImpl.java:504)
worker_1           | 	at org.newsclub.net.unix.AFUNIXSocketImpl$Lenient.setOption(AFUNIXSocketImpl.java:568)
worker_1           | 	at java.net.Socket.setSoTimeout(Socket.java:1159)
worker_1           | 	at okhttp3.internal.connection.RealConnection.connectSocket(RealConnection.java:261)
worker_1           | 	at okhttp3.internal.connection.RealConnection.connect(RealConnection.java:183)
worker_1           | 	at okhttp3.internal.connection.ExchangeFinder.findConnection(ExchangeFinder.java:224)
worker_1           | 	at okhttp3.internal.connection.ExchangeFinder.findHealthyConnection(ExchangeFinder.java:108)
worker_1           | 	at okhttp3.internal.connection.ExchangeFinder.find(ExchangeFinder.java:88)
worker_1           | 	at okhttp3.internal.connection.Transmitter.newExchange(Transmitter.java:169)
worker_1           | 	at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:41)
worker_1           | 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:142)
worker_1           | 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:117)
worker_1           | 	at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.java:94)
worker_1           | 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:142)
worker_1           | 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:117)
worker_1           | 	at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.java:93)
worker_1           | 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:142)
worker_1           | 	at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.java:88)
worker_1           | 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:142)
worker_1           | 	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:117)
worker_1           | 	at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:229)
worker_1           | 	at okhttp3.RealCall$AsyncCall.execute(RealCall.java:172)
```

pom.xml
```diff
@@ -170,6 +171,11 @@ under the License.
                             </configuration>
                         </execution>
                     </executions>
+                    <configuration>
```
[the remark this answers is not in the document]
slinkydeveloper: Ops i wrongly committed that. On my machine Fedora 32 without that config it doesn't build

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/httpfn/HttpFunctionProvider.java
slinkydeveloper: Ah, that's interesting... So I suppose we need a placeholder value like https://github.com/square/okhttp/blob/master/samples/unixdomainsockets/src/main/java/okhttp3/unixdomainsockets/ClientAndServer.java#L48

slinkydeveloper: Ok I changed the invocation of `HttpUrl.get` to use the `String` override, where the documentation states

```
@throws IllegalArgumentException If {@code url} is not a well-formed HTTP or HTTPS URL.
```

Hopefully this should avoid silent errors

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/httpfn/OkHttpUtils.java
slinkydeveloper: Ah i see:

```
  static final List<Protocol> DEFAULT_PROTOCOLS = Util.immutableList(
      Protocol.HTTP_2, Protocol.HTTP_1_1);
```

I'm gonna remove this

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/httpfn/HttpFunctionProvider.java
slinkydeveloper: Ok i'm on it

slinkydeveloper: Ok It looks like the socket factory doesn't really work as we expect...

igalshilman: In that case I can see two options:
* try out jnrsockets as provided in the original example.
* provide a new RequestReplyClient that is specialized for Unix domain socket, and backed by something else than okhttp.

slinkydeveloper: Opened an issue on junixsocket https://github.com/kohlschutter/junixsocket/issues/79

slinkydeveloper: Looking at the activity on the two projects, to me it looks a little bit more mantained/mature junixsocket. 
Let's see if the solution with junixsocket is trivial, otherwise I'll switch to jnrsockets.

igalshilman:
Hi @slinkydeveloper,
I've been investigating this and came up with a working solution:
1) basically we need to wrap the `AFUnixSocket` with a `Socket` and delegate selected operations. We can skip the socket option configuration because we set a timeout for the entire request.
2) we need to provide a Dns resolver that basically returns the same `InetAddress`.

I was able to do it, on top of your PR. 
would you like me to complete this over my branch or would you like to continue working on that?

slinkydeveloper:
> would you like me to complete this over my branch or would you like to continue working on that?

I think it's more practical if you continue to work on that and then you loop me in the review to merge your branch in master

tzulitai:
Closing. Resolved via #114 (1c018e19d96f44334fe1b24280337679dc2a7be8).

PR #116, opened 2020-05-27
[FLINK-17875] Initial support of state TTL for remote functions

tzulitai:
This is an initial support for state TTL for remote functions.

Users define state TTL in their YAML modules like so:
```
functions:
  - function:
     states:
       - name: xxxx
         ttlDurationMillis: 60000 # optional key
```

The current implementation has some limitations due to how we are multiplexing remote function's user state in a single `PersistedTable` (see FLINK-17954):
- The actual TTL being set will be the longest duration across all registered state
- The only supported expiration mode now is `AFTER_READ_AND_WRITE`. That can be added as a `ttlExpireMode` key in the YAML spec in the future.

## Main change log

- 8557fba to d27b143 some pre-req refactoring of `JsonModule` and `HttpFunctionSpec`. This refactors out version-specific JSON parsing logic to a separate class (`JsonModuleSpecParser`), and any concerns such as default values for optional keys, to `HttpFunctionSpec`.

- fcc25ec to 91e22ef Introduce a `StateSpec` model class to `HttpFunctionSpec` that captures the configured state ttl duration for each state. The `StateSpec` is then used in `RequestReplyFunction` to correctly set the state TTL on the state handles.

- 39a8745 to 0f7eb9a Adds a spec parser for v2.0 format, and wires it in to allow the version to be recognized.

- f48c64e Refactors `JsonModuleTest` so that it runs tests for both v1.0 and v2.0 format

## Verifying

- New `JsonModuleTest` covers basic parsing of the new format
- End-to-end tests don't fail imply that remote function state access is not broken due to this change
- Manually adapted the Python greeter example to check that state is expired.


tzulitai:
I'll close this, and have 2 new PRs -
one with the new refactoring approach we discussed offline, and another one on top of what which includes the actual changes for the state TTL.

PR #117, opened 2020-05-27
[FLINK-17690] Python function wrapper omits docstr

abc863377:
Use decorator to avoid losing docstr.


igalshilman:
I don’t understand this change, can you please revisit the problem description specified in the JIRA issue?

tzulitai:
I'm closing this due to inactivity, and the fact that the proposed PR does not solve the actual problem. @abc863377 please reopen if you plan to still work on this, thank you!

PR #118, opened 2020-05-27
[FLINK-17963] [core] Revert execution environment patching in StatefulFunctionsJob

tzulitai:
This is a revert of FLINK-16926, which was a temporary workaround for the problem back in Flink 1.10.0 where values present in flink-conf.yaml were not being respected via the StreamPlanEnvironment, which was used when using a JobClusterEntrypoint to start the job.

## Verifying

The end-to-end tests cover this, as they use flink-conf.yaml files with checkpointing configuration set. If the configuration were not being respected, then the exactly-once E2E test would fail.


tzulitai:
cc @igalshilman

tzulitai:
Also manually verified that with this revert, settings in `flink-conf.yaml` are being picked up.

Merging this ...

PR #120, opened 2020-05-28
[FLINK-17875] [core] State TTL for remote functions

tzulitai:
This PR is based on the refactoring work in #119, and completes the initial support for state TTL in remote functions. Only last 5 commits are relevant.

Users define state TTL in their YAML modules like so:
```
functions:
  - function:
     states:
       - name: xxxx
         expireAfter: 60000millisecond or 5sec / etc. # optional key
```

The current implementation has some limitations due to how we are multiplexing remote function's user state in a single PersistedTable (see FLINK-17954):

- The actual TTL being set will be the longest duration across all registered state
- The only supported expiration mode now is `AFTER_READ_AND_WRITE`. That can be added as a `expireMode` key in the YAML spec in the future.

## Brief change log

- 1dbfa07 to 199d3fd Introduces a `StateSpec` class which captures the state name and configured expiration timeout.
- 435060f Updates the `FunctionJsonEntity` to recognize the new format (for version 2.0)
- fb76fc5 Parameterize `JsonModuleTest` so that it runs tests for both v1.0 and v2.0 formats.

## Verifying

- The parameterized `JsonModuleTest` should cover this change.
- Manually verified functionality by locally running a modified version of the Python greeter example, with state TTL enabled.


igalshilman:
Thanks Gordon, the refactoring is already a nice improvement over the previous JsonModel, and exposing state expiration in the remote api is very helpful!

Thanks and LGTM

tzulitai:
Thanks for the review!
Merging this now ...

PR #123, opened 2020-06-01
[FLINK-18015] [doc] Add documentation for state expiration

tzulitai:
cc @sjwiesman for review.

### DIFF
--- a/docs/sdk/java.md
+++ b/docs/sdk/java.md
@@ -364,6 +364,47 @@ PersistedTable<String, Integer> table = PersistedTable.of("my-table", String.cla
 PersistedAppendingBuffer<Integer> buffer = PersistedAppendingBuffer.of("my-buffer", Integer.class);
 {% endhighlight %}
 
+### State Expiration
+
+Persisted states may be configured to expire and be deleted after a specified duration.
+This is supported by all types of state:
+
+{% highlight java %}
+@Persisted
+PersistedValue<Integer> table = PersistedValue.of(
+    "my-value",
+    Integer.class,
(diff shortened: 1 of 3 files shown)


docs/sdk/java.md
```diff
@@ -364,6 +364,47 @@ PersistedTable<String, Integer> table = PersistedTable.of("my-table", String.cla
...
+
+{% highlight java %}
+Expiration.expireAfterWriting(...)
+
+Expiration.expireAfterReadingOrWriting(...)
+{% endhighlight %}
+
+The above modes differ in when expiration timers for the state will be registered, either only on writes or on both reads and writes.
+
+These timers will also be persisted as managed state. This allows the timers to still be effectively running while the application is down.
```
tzulitai, line 403: I would need some suggestions from a native speaker here.
Not sure if this sentence reads well at all.

docs/sdk/python.md
```diff
@@ -158,9 +158,8 @@ All stateful functions may contain state by merely storing values within the ``c
 The data is always scoped to a specific function type and identifier.
 State values could be absent, ``None``, or a ``google.protobuf.Any``.
 
-<div class="alert alert-info">
```
tzulitai, line 161: Note: I removed these tags because they were causing the link for `Remote modules` to not be rendering correctly.

docs/sdk/java.md
sjwiesman: > State TTL configurations are made fault-tolerant by the runtime. In the case of downtime, state entries that should have been removed during said downtime will be purged immediately on restart.

sjwiesman:
One suggestion, otherwise +1

PR #124, opened 2020-06-01
[FLINK-18016] [doc] Add documentation for UDS domain socket transport

tzulitai:
Based on #123 just to avoid any merge conflicts in the docs.

cc @sjwiesman @igalshilman

### DIFF
--- a/docs/sdk/java.md
+++ b/docs/sdk/java.md
@@ -364,6 +364,47 @@ PersistedTable<String, Integer> table = PersistedTable.of("my-table", String.cla
 PersistedAppendingBuffer<Integer> buffer = PersistedAppendingBuffer.of("my-buffer", Integer.class);
 {% endhighlight %}
 
+### State Expiration
+
+Persisted states may be configured to expire and be deleted after a specified duration.
+This is supported by all types of state:
+
+{% highlight java %}
+@Persisted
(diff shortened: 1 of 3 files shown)


docs/sdk/java.md
```diff
@@ -364,6 +364,47 @@ PersistedTable<String, Integer> table = PersistedTable.of("my-table", String.cla
...
+
+{% highlight java %}
+Expiration.expireAfterWriting(...)
+
+Expiration.expireAfterReadingOrWriting(...)
+{% endhighlight %}
+
+The above modes differ in when expiration timers for the state will be registered, either only on writes or on both reads and writes.
+
+These timers will also be persisted as managed state. This allows the timers to still be effectively running while the application is down.
```
igalshilman, line 403: Imho this is an implemention detail, and is it true for RocksDB? I thought state is TTLed by background compaction.
My suggestion is to remove that paragraph, but let’s see what Seth thinks.

sjwiesman: Agreed, I put my suggestion on the other PR.

tzulitai: I was using the term timers here generically, just to not leak too much implementation details about how the state backends actually persist registered timers.
But I agree, Seth's suggestion in the other PR looks good, will apply that and merge!

PR #126, opened 2020-06-16
[FLINK-17954] Demultiplex remote function state

tzulitai:
This PR sits on top of #125, to make use of the new `PersistedStateRegistry` primitive.
Only the last 3 commits are relevant.

It de-multiplexes the `PersistedTable` used to hold all state values of remote functions, into separate `PersistedValue`s. This opens up each declared remote state to have their own configuration, such as state expiration.

This PR also provides a backwards compatibility path for users migrating from <= 2.1.x.
The migration path is not enabled by default; users would have to explicitly enable it themselves on the first restore by setting `statefun.remote.migrate-legacy-state` to true.

## Brief change log

- 414e446 Adds a new abstraction `PersistedRemoteFunctionValues` that encapsulates operations around `PersistedValue`s of remote functions.
- 48dc474 Uses `PersistedRemoteFunctionValues` inside the `RequestReplyFunction`. This demultiplexes the state.
- ee3eb3c Provides a migration path to de-multiplex legacy state on restore from versions <= 2.1.0.

## Verifying

I've manually verified this by manually restoring from savepoint of the Python greeter example, taken with version 2.1.0.

There is also existing unit test coverage in the `RequestReplyFunctionTest`.


statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/state/StateBinder.java
```diff
@@ -39,28 +37,22 @@ public StateBinder(@Label("state") State state) {
     this.state = Objects.requireNonNull(state);
   }
 
-  public void bind(FunctionType functionType, @Nullable Object instance) {
```
igalshilman: 👍

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/state/PersistedStates.java
```diff
@@ -26,14 +26,31 @@
...
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import org.apache.flink.statefun.sdk.FunctionType;
 import org.apache.flink.statefun.sdk.annotations.Persisted;
 import org.apache.flink.statefun.sdk.state.PersistedAppendingBuffer;
 import org.apache.flink.statefun.sdk.state.PersistedTable;
 import org.apache.flink.statefun.sdk.state.PersistedValue;
 
-final class PersistedStates {
+public final class PersistedStates {
```
igalshilman, line 37: I think that this class should reflect the fact that it reflectively looks for states.
i.e. `ReflectiveStateBinder`.

statefun-flink/statefun-flink-core/src/test/java/org/apache/flink/statefun/flink/core/state/PersistedStatesTest.java
```diff
@@ -96,6 +97,24 @@ public void bindPersistedAppendingBuffer() {
     assertThat(state.boundNames, hasItems("buffer"));
   }
 
+  @Test
+  public void bindDynamicState() {
```
igalshilman, line 101: 👍 that is a super doper great test

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/reqreply/PersistedRemoteFunctionValues.java
```diff
@@ -0,0 +1,76 @@
...
+
+  private final Map<String, PersistedValue<byte[]>> managedStates;
+
+  public PersistedRemoteFunctionValues(List<StateSpec> stateSpecs) {
+    Objects.requireNonNull(stateSpecs);
+    this.managedStates = new HashMap<>(stateSpecs.size());
+    stateSpecs.forEach(spec -> managedStates.put(spec.name(), createStateHandle(spec)));
+  }
+
+  void forEach(BiConsumer<String, byte[]> consumer) {
```
igalshilman, line 45: since this is very tied to the `ToFunction` aspect of the code, can we pass in the `ToFunction` builder to avoid a allocation for the closure?

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/reqreply/RequestReplyFunction.java
```diff
@@ -228,27 +204,27 @@ private void handleOutgoingDelayedMessages(Context context, InvocationResponse i
...
-      ToFunction.PersistedValue.Builder valueBuilder =
-          ToFunction.PersistedValue.newBuilder().setStateName(stateSpec.name());
-
-      byte[] stateValue = managedStates.get(stateSpec.name());
-      if (stateValue != null) {
-        valueBuilder.setStateValue(ByteString.copyFrom(stateValue));
-      }
-      batchBuilder.addState(valueBuilder);
-    }
+    managedStates.forEach(
```
igalshilman, line 207: I think that mangedState should have a `populateBatch` method,
to avoid the forEach allocation.

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/state/PersistedStates.java
tzulitai: I'd like to avoid naming this class a `*StateBinder` as well, to differentiate it from an actual `StateBinder` implementation.

I chose to rename the static method to `findReflectivelyAndBind` instead.

PR #127, opened 2020-06-30
[hotfix] [docs] Fix typos

uce:
Really helpful documentation! Yesterday, I read over most of the docs and found a few typos, including some typos in code. Feel free to push back on any of the changes that seem unnecessary to you.

A few nit comments:
* In the [project setup](https://github.com/apache/flink-statefun/blob/8376afa/docs/getting-started/project-setup.md#build-project) page, we state that ``mvn clean package`` produces an uber JAR which is not true. Is the intention to provide a Maven configuration that does create an uber JAR or are the docs mistaken?
* In the [Python walk-through](https://github.com/apache/flink-statefun/blob/8376afa/docs/getting-started/python_walkthrough.md#our-first-function), we first write a function without a known type and then show how to specify a type. I'm wondering we should only show the known type example and keep it as a note that it's also possible to the use the `Any` type.
* Is RocksDB a strict requirement or can users configure any state backend? If RocksDB is a requirement, we could consider pointing this out in https://github.com/apache/flink-statefun/blob/8376afa/docs/concepts/logical.md#function-lifecycle (last sentence).
* In the Java SDK, we mix tab and space for code formatting. I think spaces are easier to read in the docs (I understand that the code should follow the apache/flink style guide).
* In the [Java SDK async requests section](https://github.com/apache/flink-statefun/blob/8376afa/docs/sdk/java.md#completing-async-requests), we could use a stateful match function to reduce some of the boilerplate.
* Overall, message routing still felt a bit mysterious after reading over the docs. I don't know if others agree, but we might want to add more details on this part of the system.

### DIFF
--- a/docs/deployment-and-operations/packaging.md
+++ b/docs/deployment-and-operations/packaging.md
@@ -38,7 +38,7 @@ The provided base image allows teams to package their applications with all the 
 Below is an example Dockerfile for building a Stateful Functions image with both an [embedded module]({{ site.baseurl }}/sdk/modules.html#embedded-module) and a [remote module]({{ site.baseurl }}/sdk/modules.html#remote-module) for an application called ``statefun-example``.
 
 {% highlight dockerfile %}
-FROM flink-statefun:{{ site.version }}
+FROM ververica/flink-statefun:{{ site.version }}
 
 RUN mkdir -p /opt/statefun/modules/statefun-example
 RUN mkdir -p /opt/statefun/modules/remote
--- a/docs/deployment-and-operations/packaging.md
+++ b/docs/deployment-and-operations/packaging.md
@@ -50,7 +50,7 @@ COPY module.yaml /opt/statefun/modules/remote/module.yaml
 {% if site.is_stable %}
 <div class="alert alert-info">
 	The Flink community is currently waiting for the official Docker images to be published to Docker Hub.
-	In the meantime, Ververica has volunteered to make Stateful Function's images available via their public registry: 
+	In the meantime, Ververica has volunteered to make Stateful Functions' images available via their public registry: 
 
 	<code class="language-dockerfile" data-lang="dockerfile">
 		<span class="k">FROM</span><span class="s"> ververica/flink-statefun:{{ site.version }}</span>
--- a/docs/deployment-and-operations/packaging.md
+++ b/docs/deployment-and-operations/packaging.md
@@ -38,7 +38,7 @@ The provided base image allows teams to package their applications with all the 
 Below is an example Dockerfile for building a Stateful Functions image with both an [embedded module]({{ site.baseurl }}/sdk/modules.html#embedded-module) and a [remote module]({{ site.baseurl }}/sdk/modules.html#remote-module) for an application called ``statefun-example``.
 
 {% highlight dockerfile %}
-FROM ververica/flink-statefun:{{ site.version }}
+FROM flink-statefun:{{ site.version }}
 
 RUN mkdir -p /opt/statefun/modules/statefun-example
 RUN mkdir -p /opt/statefun/modules/remote
--- a/docs/getting-started/python_walkthrough.md
+++ b/docs/getting-started/python_walkthrough.md
@@ -170,16 +170,16 @@ from statefun import StatefulFunctions
 functions = StatefulFunctions()
 
 @functions.bind("example/greeter")
-def greet(context, message: GreetRequest):
+def greet(context, greet_request: GreetRequest):
     response = GreetResponse()
-    response.name = message.name
-    response.greeting = "Hello {}".format(message.name)
+    response.name = greet_request.name
+    response.greeting = "Hello {}".format(greet_request.name)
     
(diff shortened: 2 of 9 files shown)


docs/deployment-and-operations/packaging.md
```diff
@@ -38,7 +38,7 @@ The provided base image allows teams to package their applications with all the
 Below is an example Dockerfile for building a Stateful Functions image with both an [embedded module]({{ site.baseurl }}/sdk/modules.html#embedded-module) and a [remote module]({{ site.baseurl }}/sdk/modules.html#remote-module) for an application called ``statefun-example``.
 
 {% highlight dockerfile %}
-FROM flink-statefun:{{ site.version }}
+FROM ververica/flink-statefun:{{ site.version }}
```
tzulitai, line 41: I'd prefer not to change the base image mentioned here, as the `ververica/` hosted one is considered a temporary solution for the time being. We do mention the Ververica image as a side note below, so this should be fine?

docs/getting-started/python_walkthrough.md
```diff
@@ -338,7 +338,7 @@ The base image provides the Stateful Function runtime, it will use the provided
 This can be found in the `Dockerfile` in the root directory. 
 
 {% highlight docker %}
-FROM flink-statefun:{{ site.version }}
+FROM ververica/flink-statefun:{{ site.version }}
```
tzulitai, line 341: Same as comment above.

docs/deployment-and-operations/packaging.md
uce: I agree that `flink-statefun` without the `ververica` repository is preferrable, but I think we haven't published the image as an official library image yet (at least I can't `docker pull flink-statefun:2.1.0`), right?

My reasoning for the change was that with the `ververica` repository prefix the Dockerfile snippets work without requiring users to manually build the statefun image locally. What do you think?

(I'm happy to revert this change. I actually kept it as a separate commit, because I was expecting that this might have been on purpose.)

sjwiesman: I agree with @tzulitai in principle, but it doesn't look like there is any timeline on getting the official docker images out. I think we add the ververica registry for now, we can always retroactivly remove it from old versions of the docs after they are published.

uce: So, :shipit:?

igalshilman: Maybe that should be brought up to a vote/discussion in the mailing list? what do you think @uce can you raise that issue?
My personal opinion here, that currently it is not a nice experience for anyone how would like to try the project and I don't see any progress in https://github.com/docker-library/official-images/pull/7749

docs/getting-started/python_walkthrough.md
```diff
@@ -170,16 +170,16 @@ from statefun import StatefulFunctions
 functions = StatefulFunctions()
 
 @functions.bind("example/greeter")
-def greet(context, message: GreetRequest):
+def greet(context, greet_request: GreetRequest):
```
igalshilman, line 173: Thanks for suggesting a more readable snippet 👍

docs/sdk/java.md
```diff
@@ -162,7 +162,7 @@ Finally, if a catch-all exists, it will be executed or an ``IllegalStateExceptio
 
 ## Function Types and Messaging
 
-In Java, function types are defined as a _stringly_ typed reference containing a namespace and name.
+In Java, function types are defined as a _strongly_ typed reference containing a namespace and name.
```
igalshilman, line 165: It is `stringly typed` right @sjwiesman ?

sjwiesman: I used stringly on purpose because you aren't refereing to the class but a string "<namespace>/<name>". I don't think strongly is correct but we can change it. How about, 

```
In Java, function types are defined as logical pointers composed of a namespace and name
```
?

uce: Yes, my suggestion didn't make sense. +1 to change it as Seth outlined.

"stringly typed" has a negative connotation for me as in [needlessly using strings when other options are available](https://wiki.c2.com/?StringlyTyped).

docs/deployment-and-operations/packaging.md
uce: I had pinged Igal offline to start the discussion, because he was involved in the migration away from Ververica.

Since this is controversial, I will revert the commit and hope that we can merge this.

docs/getting-started/python_walkthrough.md
uce: Reverted.

docs/deployment-and-operations/packaging.md
uce: Reverted.

docs/sdk/java.md
uce: Applying Seth's suggestion.

uce:
@sjwiesman @igalshilman @tzulitai I reverted the Docker image repository change and applied Seth's suggestion for the function types. Can you take another look and merge or close this PR?

tzulitai:
Thanks @uce, will merge this!

PR #132, opened 2020-08-17
[FLINK-18979][docs] Update docs to better emphasize remote functions

sjwiesman:


docs/index.md
```diff
@@ -23,16 +23,16 @@ specific language governing permissions and limitations
 under the License.
 -->
 
-**Stateful Functions** is an open source framework that reduces the complexity of building and orchestrating distributed stateful applications at scale.
-It brings together the benefits of stream processing with Apache Flink® and Function-as-a-Service (FaaS) to provide a powerful abstraction for the next generation of event-driven architectures.
+**Stateful Functions** is a framework, developed under the umbrella of Apache Flink, that makes it simple to build **consistent stateful serverless applications**.
+It is designed to work with modern architectures, like stateless Kubernetes deployments or popular event-driven FaaS platforms like AWS Lambda, KNative, etc.,
```
sjwiesman, line 27:
```suggestion
It is designed to work with modern architectures, like cloud-native deployments or popular event-driven FaaS platforms like AWS Lambda, KNative, etc.,
```

This was my other idea for this sentence, I'd like to hear what people like more

docs/index.md
```diff
@@ -23,16 +23,16 @@ specific language governing permissions and limitations
 under the License.
 -->
 
-**Stateful Functions** is an open source framework that reduces the complexity of building and orchestrating distributed stateful applications at scale.
-It brings together the benefits of stream processing with Apache Flink® and Function-as-a-Service (FaaS) to provide a powerful abstraction for the next generation of event-driven architectures.
+**Stateful Functions** is a framework, developed under the umbrella of Apache Flink, that makes it simple to build **consistent stateful serverless applications**.
```
sjwiesman, line 26:
```suggestion
**Stateful Functions** is a framework, developed under the umbrella of Apache Flink, that simplifies building distributed stateful applications.
```

Other possible first sentence

docs/index.md
tzulitai: Stateless k8s deployments, IMO, would probably catch people a bit confused from a first read.
Overall, I always had the feeling that due to how the project is named "Stateful Functions", every time we use the term "stateless" we are in a bit of risk of misguiding readers who don't fully understand the context the term is used in.

docs/index.md
tzulitai: I personally like the first version more, and have been using that tag line a lot in recent talks / current work-in-progress blog posts. However, I'm also not sure if the first version is "down-scoping" the project as well.

I'd leave this to @igalshilman / @StephanEwen though, as it seems like more a question of how we want the project to be pitched moving forward.

tzulitai: Just my two more cents:
From how I read the changes in the PR, it seems like the overall goal is to make remote functions / the "Flink as DB" mode more prominent and the main focus of the project.

In that sense, I would vote for `**... makes it simple to build consistent stateful serverless applications` as that seems to be more coherent with the goal.

tzulitai:
While you're at it, once we've reached consensus of the updated contents of the homepage, could you update the README.mds (at root and under `statefun-python-sdk`) as well?

docs/index.md
```diff
@@ -23,16 +23,16 @@ specific language governing permissions and limitations
...
 <p class="text-center">
-  <img width="80%" src="{{ site.baseurl }}/fig/stateful_functions.png"/>
+    <img width="80%" src="{{ site.baseurl }}/fig/concepts/arch_overview.svg"/>
 </p>
 
-There are many ways to get started with Stateful Functions.
-Which one is the best for you depends on your goals and prior experience.
-Whether you prefer a more theoretical or a practical approach, we hope you’ll find this section helpful.
+It brings together the benefits of Apache Flink®, streaming semantics for processing large datasets with low latency and bounded resource constraints,
+along with a runtime for modeling stateful entities that supports location transparency, concurrency, scaling, and resiliency.
```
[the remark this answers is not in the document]
sjwiesman: -1 I think its "resiliency"

morsapaes:
If we're changing the docs, should we also change the webpage in the Flink website?

sjwiesman:
I've pushed an update to the home page. @tzulitai Yes I will update the README's after we reach consensus.

sjwiesman:
@morsapaes yes, we can as a follow up.

sjwiesman:
@tzulitai / @morsapaes when you get a chance I've updated the homepage based on the above discussion.

tzulitai:
LGTM, merging. Thanks @sjwiesman!

PR #135, opened 2020-08-31
[FLINK-19095] [remote] Allow state expiration mode for remote function state

tzulitai:
Before this PR, since state was multiplexed, we didn't allow state to have individual state expiration modes, and everything was defaulted to expire with "after-read-and-write" mode.

Now with state de-multiplexed, it is possible to allow this, with the following configuration:

```
functions:
    function:
        spec:
            states:
                - name: xxxx
                  expireAfter: 60000millisecond
                  expireMode: [after-invoke | after-write]
```

Note that I named the usual `after-read-or-write` module as `after-invoke` in the remote function case.
This is due to the fact that they are semantically equivalent in the remote function case, and naming it as such seems to be less confusing for users using remote functions.


igalshilman (approved):
Thanks for adding this,
LGTM!

PR #136, opened 2020-08-31
[FLINK-19096] [sdk] Rework PersistedStateRegistry registration methods

tzulitai:
This reworks the `registerValue` / `registerTable` etc. methods on `PersistedStateRegistry` to directly accept a constructed `PersistedValue` / `PersistedTable` instance.
This allows us to avoid having to synchronize the arguments of those methods and the state class constructors.

The usage now looks like this:

```
public class MyFunction implements StatefulFunction {
 
      @Persisted
      PersistedStateRegistry registry = new PersistedStateRegistry();
 
      public MyFunction() {
            // works during function instantiation ...
            registry.registerValue(PersistedValue.of(...));
      }

      @Override
      public void invoke(Context context, Object input) {
          // or during runtime
          registry.registerValue(PersistedValue.of(...));
      }
 }
```

The `register*` methods throw if a duplicate registration occurred for the same state name.

## Validating the change

All existing tests in `PersistedStateRegistryTest` and `PersistedStatesTest` cover this change.


igalshilman (approved):
Looks good, thanks for improving this!

PR #137, opened 2020-08-31
[FLINK-19102] [core, sdk] Make StateBinder a per-FunctionType entity

tzulitai:
This PR is based on top of #136 because it touches a few overlapping classes.
Only last 2 commits are relevant.

---

Prior to this PR, a single `StateBinder` instance handles binding state objects to Flink state for multiple `FunctionType`s.

With these changes, this is changed so that each `FunctionType` has its own `StateBinder` (specifically, a `FlinkStateBinder`).
This allows a cleaner separation of concerns in the `PersistedStateRegistry`, so that the information about which function type the registry is bounded is not leaked into the SDK classes, but instead encapsulated within the `StateBinder` being used at runtime.

---

## Verifying this change

The existing tests in `PersistedStatesTest` cover the changes introduced by this refactoring.

### DIFF
--- a/statefun-sdk/src/main/java/org/apache/flink/statefun/sdk/state/PersistedAppendingBuffer.java
+++ b/statefun-sdk/src/main/java/org/apache/flink/statefun/sdk/state/PersistedAppendingBuffer.java
@@ -160,6 +160,13 @@ public final class PersistedAppendingBuffer<E> {
     accessor.clear();
   }
 
+  @Override
+  public String toString() {
+    return String.format(
+        "PersistedAppendingBuffer{name=%s, elementType=%s, expiration=%s}",
+        name, elementType.getName(), expiration);
+  }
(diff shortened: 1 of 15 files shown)


statefun-sdk/src/main/java/org/apache/flink/statefun/sdk/state/PersistedAppendingBuffer.java
```diff
@@ -160,6 +160,13 @@ public void clear() {
     accessor.clear();
   }
 
+  @Override
```
igalshilman, line 163: I think that it would be somewhat not intuitive for users not to see the actual value, what do you think?

igalshilman (approved):
The last two commits LGTM, not having to specify the function type in the dynamic state registration is a nice improvement.

PR #138, opened 2020-09-01
[FLINK-19106] [core] Add more timeout configs for remote functions

tzulitai:
Prior to this PR, we only support setting the call timeout for remote functions, which spans a complete call including connection, writing request, server-side processing, and reading response.

To allow more fine-grained control of this, this PR introduces configuration keys for `connectTimeout` / `readTimeout` / `writeTimeout` to remote function specs.
By default, these values should be 10  seconds to be coherent with the current behaviour

Usage:
```
function:
      meta:
            kind: http
            type: com.foo/world
      spec:
            endpoint: http://localhost:5959/statefun
            states:
                ...
            timeout: 2minutes
            connectTimeout: 20seconds
            readTimeout: 1second
            writeTimeout: 10seconds
```

### DIFF
--- a/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/httpfn/HttpFunctionSpec.java
+++ b/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/httpfn/HttpFunctionSpec.java
@@ -31,12 +31,18 @@ public final class HttpFunctionSpec implements FunctionSpec, Serializable {
   private static final long serialVersionUID = 1;
 
   private static final Duration DEFAULT_HTTP_TIMEOUT = Duration.ofMinutes(1);
+  private static final Duration DEFAULT_HTTP_CONNECT_TIMEOUT = Duration.ofSeconds(10);
+  private static final Duration DEFAULT_HTTP_READ_TIMEOUT = Duration.ofSeconds(10);
+  private static final Duration DEFAULT_HTTP_WRITE_TIMEOUT = Duration.ofSeconds(10);
   private static final Integer DEFAULT_MAX_NUM_BATCH_REQUESTS = 1000;
 
   private final FunctionType functionType;
   private final URI endpoint;
   private final List<StateSpec> states;
   private final Duration maxRequestDuration;
+  private final Duration connectTimeout;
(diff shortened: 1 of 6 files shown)


statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/httpfn/HttpFunctionSpec.java
```diff
@@ -111,14 +138,36 @@ public Builder withMaxRequestDuration(Duration duration) {
...
+
     public Builder withMaxNumBatchRequests(int maxNumBatchRequests) {
       this.maxNumBatchRequests = maxNumBatchRequests;
       return this;
     }
 
     public HttpFunctionSpec build() {
       return new HttpFunctionSpec(
-          functionType, endpoint, states, maxRequestDuration, maxNumBatchRequests);
+          functionType,
```
igalshilman, line 163: I think that somewhere we should validate the timeouts that they make sense.
For example: all the durations are non zero, and the sum of read write and connect is <= call timeout.

statefun-flink/statefun-flink-core/src/test/resources/module-v2_0/module.yaml
```diff
@@ -37,6 +37,10 @@ module:
               - name: seen_count
                 expireAfter: 60000millisecond
                 expireMode: after-invoke
+            timeout: 2minutes
+            connectTimeout: 20seconds
+            readTimeout: 1second
```
igalshilman, line 42: Can the example say 10second on the read timeout?
The reason is that the examples are usually copy-pasted.

igalshilman (approved):
Thanks, the changes looks good to me!
One thing I’d recommend adding is a very basic validation of the duration values, other than that good to merge!

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/httpfn/HttpFunctionSpec.java
tzulitai: 0 duration would mean that no timeout is imposed. Should that really be an invalid configuration?

PR #139, opened 2020-09-01
[FLINK-19107] Add checkpointing and recovery options to template flink-conf.yaml

tzulitai:
A corresponding change for the template flink-conf.yaml used in distributed Docker images can be found at https://github.com/apache/flink-statefun-docker/pull/6.

This also changes the flink-conf.yaml used in e2e tests to verify that the template does indeed work out-of-box for users.


igalshilman (approved):
LGTM!

PR #141, opened 2020-09-03
[FLINK-19129] [k8s] Update log4j-console in template Helm chart

tzulitai:
This needs to be updated since logging was changed in Flink from 1.10.x to 1.11.x.

The new contents is identical to how it is in Flink 1.11.x: https://github.com/apache/flink/blob/release-1.11/flink-dist/src/main/flink-bin/conf/log4j-console.properties


tzulitai:
cc @igalshilman

PR #142, opened 2020-09-04
[FLINK-19130] [core] Expose metrics for backpressure

tzulitai:
This PR adds in total 2 new metrics related to backpressure:
- Number of blocked addresses (per function type)
- Number of inflight async operations (per function type + per operator)

We also rejected to add the following metric, since after some discussion it doesn't seem to add much value:
- Number of accumulated records pre-flight in batch per function type.
This was not added, with the assumption that users would really only want to care about that some address has reached the maximum request batch size and was being blocked.

---

## Verification

I verified this by running the Python Greeter example, with the following modifications to let backpressure happen more easily:
- Maximum batch size = 1
- No delay between each generated message to have maximum input rate

You can see the following metric charts in the Flink Web UI:

image
 
---

## Brief changelog

- 78cbc19 Extends the `FunctionTypeMetrics` interface to include the new metrics, and adds a new `FunctionDispatcherMetrics` interface for per-operator metrics.
- 6408866 Introduce a scoped-down interface `FunctionTypeMetricsRepository` and let `StatefulFunctionsRepository` extend it. Components that need to access function metrics will be passed this interface.
- 0fbb17a preliminary extension to `ObjectContainer` DI utility so that we can share same instance across different object labels.
- 6c77ca4 Wire-in the new metrics in `AsyncSink` to expose backpressure metrics.

### DIFF
--- a/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/backpressure/BackPressureValve.java
+++ b/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/backpressure/BackPressureValve.java
@@ -53,4 +53,12 @@ public interface BackPressureValve {
    * @param address the address
    */
   void blockAddress(Address address);
+
+  /**
+   * Checks whether a given address was previously blocked with {@link #blockAddress(Address)}.
+   *
+   * @param address the address to check
+   * @return boolean indicating whether or not the address was blocked.
+   */
+  boolean isAddressBlocked(Address address);
 }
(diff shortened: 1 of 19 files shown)


tzulitai:
Discussion point:

Strictly speaking, the per-operator `inflight-async-ops` metric is redundant, since most metric reporter systems commonly used by Flink (Prometheus / InfluxDB etc.) have a query language for aggregating metrics.
Therefore, the per-operator `inflight-async-ops` metric can be aggregated from the per-function ones.

@igalshilman what do you think about this?

tzulitai:
Update:
After another offline discussion with @igalshilman, we decided to still add metrics for backlog monitoring, per function type.
It's a rather low-hanging fruit to add, and would serve as a good starter for further exploring what auto-scalers might need.

igalshilman (approved):
Thanks @tzulitai, LGTM!

PR #148, opened 2020-09-11
[FLINK-19199] [core] Add execution attempt ID to feedback channel keys

tzulitai:
The feedback brokers identify a single pair of producer / consumer with a `SubtaskFeedbackKey`, which is uniquely identified within a job by only the subtask index.

This can potentially become an issue in job restart scenarios, since the feedback brokers are static fields, and are durable across non-TM failure job restarts.
Currently it is not a problem because we do clear broker channels on close, but adding an extra execution attempt ID into the `SubtaskFeedbackKey` can make things safer.

---

## Verifying this change

There are existing unit tests in `FeedbackChannelTest`.
Moreover, it is assumed that the failure recover / exactly-once E2E tests have sufficient coverage for this change.


igalshilman (approved):
LGTM! thanks.

PR #153, opened 2020-09-16
[FLINK-19256] [core] Move Flink config validation out of StatefulFunctionsConfig

tzulitai:
The motivation for this change was that in the DataStream API example, there were unnecessary configuration that existed just because of how we validate required settings when creating a `StatefulFunctionsConfig`.

However, those settings are only needed when executing a StateFun job through the `StatefulFunctionsJob` entry point.

This PR refactors the validation out of `StatefulFunctionsConfig` to only happen in `StatefulFunctionsJob`.


igalshilman (approved):
Thanks @tzulitai,
looks pretty good!

PR #160, opened 2020-09-21
[hotfix][dockerfile] Add newline to end of flink-conf

sjwiesman:
I ran into this while doing release testing, this is mostly quality of life and I wouldn't consider a release blocker. 

When building small applications it's often easy to simply append flink configurations to the provided base image directly inside the Dockerfile. echo does not prepend a newline so its convenient to leave one at the end of the file. Otherwise, the first configuration is added to the task manager memory amount which leads to a strange parse error on
deployment.

Basically, it allows support of the following:

```Dockerfile
FROM flink-statefun:2.2-SNAPSHOT
RUN echo "execution.checkpointing.interval: 5sec" >> /opt/flink/conf/flink-conf.yaml;
```


tzulitai:
+1, LGTM

PR #163, opened 2020-10-14
[FLINK-19620] Merge ExactlyOnceE2E and RemoteModuleE2E

tzulitai:
The original `ExactlyOnceE2E` verified end-to-end exactly-once with TM failures, while the `RemoteModuleE2E` verified a deployment with remote functions.

As mentioned in #159 by @sjwiesman, we seem to be lacking an exactly-once verifying E2E with remote functions.

It is actually worthwhile to merge these 2 E2Es into one, since 1) remote functions are built on top of embedded functions, so merging them does not hurt us on test coverage, and 2) E2E tests take a significant amount of test time, so less is better.

### Change log

- 012d4b3: Extends the original `RemoteModuleE2E` to have TM failures, `read_committed` settings for Kafka consumers that verify outputs, and exactly-once delivering Kafka egresses.
- 61bcfc2: Removes the `ExactlyOnceE2E`
- 7fa17b1: Renames `RemoteModuleE2E` to `ExactlyOnceWithRemoteFnE2E` to reflect the extra coverage in the test.


tzulitai:
cc @sjwiesman @igalshilman

igalshilman (approved):
LGTM 👍

PR #164, opened 2020-10-19
Fix for FLINK-19515 by moving creation of the InvocationContext out o…

fransking:
…f __init__ and into __call__.  The ic is destroyed per __call__ so needs to be instantiated per __call__ also.


igalshilman (changes requested):
Thanks for creating the PR @fransking !

The changes looks good to me, and I'm happy to merge that.
Just before merging I have two questions:
1) Were you able to verify that with this change the bug doesn't happen anymore (I'd assume you did, but just verifying)
2) Can you please amend the commit name to be of the format (you can peek at previous commits)
`[FLINK-19515][python] ...`

PR #171, opened 2020-11-05
 [FLINK-20009][docs] Add documentation link check to travis 

sjwiesman:
This adds a Travis step to verify the docs build and there are no dead links. 

I copied the `docs.sh` file directly from the main flink repo, this is how we keep the docs stable there as well.

### DIFF
--- a/tools/ci/docs.sh
+++ b/tools/ci/docs.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+################################################################################
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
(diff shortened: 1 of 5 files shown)


sjwiesman:
cc @igalshilman

tools/ci/docs.sh
```diff
@@ -0,0 +1,32 @@
...
+do
+	echo "Waiting for server..."
+	curl -Is http://localhost:4000 --fail
+	if [ $? -eq 0 ]; then
+		break
+	fi
+	sleep 10
+done
+
+./docs/check_links.sh
```
tzulitai, line 32: nit: We've made an effort to make the tool scripts in StateFun to be oblivious of the directory they're being executed from (You can also check out the release scripts under `tools/releasing` for a quick copy-paste for this). Could you also do that for this script? Or at least a sanity check that this script is being executed from the project root.

tzulitai (approved):
+1 to add this to Travis.
One nit comment below.

tools/ci/docs.sh
sjwiesman: Sounds good.

tzulitai:
+1, LGTM

PR #172, opened 2020-11-11
[FLINK-20085] Remove RemoteFunctionStateMigrator from restore code paths

tzulitai:
The `RemoteFunctionStateMigrator` was added in 2.2.x to allow migrating legacy multiplexed remote function state in versions < 2.2.0, to the new demultiplexed format starting from 2.2.

Since the 2.2 series is now released and writes only in the new formats, we can now safely remove this migration path.

What this means for users that are still on StateFun versions < 2.2.0, if they want to upgrade to 2.3.x in the future:
- They need to first upgrade to 2.2.x, and take a savepoint from that version
- Then, use the 2.2.x savepoint to upgrade to 2.3.x.

---

## Change log

- b6450d6 Remove `RemoteFunctionStateMigrator` and all its usages
- 81dd548 Remove config key `statefun.remote.migrate-legacy-state`, which was introduced in the past as an internal flag for switching on the migration path, if users require it.


igalshilman (approved):
Thanks @tzulitai, looks good to me!

PR #174, opened 2020-11-16
[FLINK-20161] Switch from Travis to Github Actions for CI

tzulitai:
Travis-CI.com recently announced a new pricing model on Nov. 2, which affects public open source projects: https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

It seems like the Travis CI jobs on `apache/flink-statefun` have already been blocked / queued indefinitely for over 2 days now because of this.

Github Actions seems to be a popular alternative nowadays:
- No limited test time with its hosted builders, if repo is public
- Activation is automatic - one step / click less for contributors to get CI running for their forks

Given that the CI workflows in `flink-statefun` is very minimal right now, we propose to make the switch to Github Actions as the efforts to do that should be relatively trivial.

**Note:** previously, the Travis CI workflow had a Java 11 build defined. That job is actually meaningless, since it is not configuring Maven to build with JDK 11.
For this reason, the migrated Github Actions workflow currently only has a job for JDK 8.

There should be a separate PR / efforts for adding in JDK 11 builds to the CI.


igalshilman (approved):
Thanks Gordon!
Looking forward of using GitHub actions.

PR #175, opened 2020-11-17
[FLINK-20189] [core] Fix Header.skipHeaderSilently to fully fill the header bytes read buffer before comparing

tzulitai:
This PR fixes the per key-group `Header.skipHeaderSilently` operation to properly fully fill the header bytes read buffer, before comparing it with the expected header bytes.

Since {{InputStream.read(byte[])}} by contract only guarantees reading at least 1 byte and could read less bytes than desired (i.e. the read buffer length), prior to this PR, it was possible that the read buffer wasn't fully filled. This would result in an invalid header bytes comparison, and incorrectly pushing back header bytes back into the remaining stream.

Ultimately, what this means is that in this case, `Header.skipHeaderSilently` doesn't actually skip the header bytes, and the following reads on the per key-group streams (to restore the feedback events) would be attempting to read a corrupt stream.

The fix is to do a `readFully` operation instead of a simple read when filling the header bytes read buffer.

## Change log

- b6256df Backports the `tryReadFully` utility in Flink. This backport can be reverted once we upgrade to Flink 1.12.0 / 1.11.3. Note that in our backport, we introduced a new `RandomReadLengthByteArrayInputStream` utility to attempt to mimic the behaviour of "real-life" input streams in our tests, where each read can return less than the desired number of bytes.

- e246c1c Use `tryReadFully` in `Header.skipHeaderSilently`.

- 099015c Extend the existing tests in `UnboundedFeedbackLoggerTest` to use the new `RandomReadLengthByteArrayInputStream`. This new version of the test would fail without applying the fix.

### DIFF
--- a/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/logger/InputStreamUtils.java
+++ b/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/logger/InputStreamUtils.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


igalshilman (approved):
Thanks @tzulitai, the changes look good to me.
As discussed offline, I'm +1 to merge after:
* remove the iterations in the repeated tests.
* add a `Precondtions.checkState` that the passed array size is larger than 0.

PR #179, opened 2020-11-26
[FLINK-20268] [python] Support new invocation protocol in the Python SDK

tzulitai:
This PR updates the Python SDK to support registered state specifications with the SDK, and implement the new protocol introduced in #177. 

Users would do the following to register state specifications for individual functions:
```
from statefun import StatefulFunctions, StateSpec, Expiration

functions = StatefulFunctions()

@functions.bind(
    typename="com.foo/bar",
    states=[
        StateSpec("state-1"),
        StateSpec("state-2", expiration=Expiration(2000, expire_mode=Expiration.Mode.AFTER_WRITE))
    ])
def fun(context, message):
    ...
```

Users would no longer register state eagerly in module YAML definitions.

User contracts:
- identical state names cannot be registered twice, otherwise a `StateRegistrationError` is raised
- if a state name was not registered, but attempted to be accessed in the function (via the context), a `StateRegistrationError` is raised

## Brief change log

- 6b3ec55 Regenerates the Python Protobuf messages due to updated Protocol.

- 44aea9c Defines the new state registration constructs for the SDK, such as `StateSpec` and `Expiration`. With this change, the `StatefulFunctions` construct now knows what state each individual function requires access to.

- bd95f71 Modify the request reply handlers to match the provided state in invocation requests against the registered state specifications of the target function. If the invocation request is missing state, the request reply handler responds with a `IncompleteInvocationContext` message. Please see the user contracts of the changes by referencing the new unit tests added in `request_reply_test.py`.

- 1d7bcd6 to dea9c2e Update all examples and E2E, and Python SDK README to reflect the new way of registering state in the Python SDK.

## Testing the change

Run Python SDK unit tests:
- Unit tests for the new `StateSpec` and `Expiration` constructs are covered in the updated `statefun_test.py`
- Unit tests for the new request reply handler logic is covered in the updated `request_reply_test.py`

E2E:
- The updated E2E `ExactlyOnceWithRemoteFnE2E` (updated to declare state in the Python functions) tests this end-to-end as a black box.

## Documentation

I've left out documentation updates in this PR, as there is an upcoming planned rework of the documentation.
Will follow-up updating the documentation separately in a new PR.


statefun-examples/statefun-python-walkthrough-example/walkthrough.py
```diff
@@ -111,12 +115,29 @@ def state2(context, message):
...
     del context['counter']
 
 
+@functions.bind(
+    typename="walkthrough/missing_state",
+    states=[
+        StateSpec("counter"),
+        StateSpec("missing-state-1"),
+        StateSpec("missing-state-2", expiration=Expiration(10000)),
+        StateSpec("missing-state-3", expiration=Expiration(2000, expire_mode=Expiration.Mode.AFTER_WRITE))
```
igalshilman, line 132: What do you think about simplifying a bit the expiration object?
having to type the `expire_mode` is a bit too much, what do you think of:
1) having separate Expiration classes for each mode.
```StateSpec("seen", expiration=AfterWrite(..))``
```StateSpec("age", expiration=AfterReadOrWrite(..)))```

2) We use Python`s `timedelta` in `send_after` what do you think of using the same thing here?
```StateSpec("age", expiration=AfterWrite(timedelta(days=3)))```

igalshilman (approved):
Thanks @tzulitai, the changes look good to me.
I've also run the greeter example with the new SDK and it works as expected.
I have one comment about the Expiration object, other than that I'm ok to merge!

statefun-examples/statefun-python-walkthrough-example/walkthrough.py
tzulitai: Very good ideas! I didn't realize the `timedelta` existed, makes sense to use that as well.

tzulitai:
Thanks for the review @igalshilman!
I've addressed your comment. Will merge this now ...

PR #181, opened 2020-11-29
[FLINK-18810][sdk] Golang remote functions SDK

sjwiesman:
Implements a remote Go SDK based on the latest invocation protocol. Tests can be executed by running `go test ./...` from within the `statefun-go-sdk` directory. There is also a greeter example you can reference. 

### Licensing  

I'm not sure how licensing works for the go sdk. We will never release binaries, go dependencies pull directly from git repositories via tags, and build locally, so we are always only providing source releases. You can check the configured dependencies in the `go.mod` file and fully resolve transitive dependencies via `go.sum` lock file. 

### Documentation 

There is go doc and a README but I'm holding off on full documentation for now. I'm assuming the API might change over the course of the review. 

### Releasing

This SDK is based on go modules, you can find full details on how to release a module [here](https://blog.golang.org/publishing-go-modules)


sjwiesman:
cc @tzulitai @igalshilman

tzulitai:
Concerning the licensing:

Since we won't be releasing artifacts for the Go SDK, I think ensuring compatibility for ASF 2.0 breaks down to:
- Are there any copied source codes? If yes, the source license should be mentioned in a NOTICE file.
- Are there any dependencies with a license not compatible with ASF 2.0?
- Are there any dependencies that are modified in a way that violates its original license (e.g. shading).

I quickly skimmed through the dependencies:

- `github.com/golang/protobuf`: BSD 3-clause
- `github.com/stretchr/testify`: MIT
- `github.com/valyala/bytebufferpool`: MIT
- `google.golang.org/protobuf`: BSD 3-clause

Transitive dependencies are neither bundled or modified / copied in any way.

Therefore, I think we should good on the licensing side ✅

sjwiesman:
closing until the new protocol is in place

PR #182, opened 2020-12-01
[FLINK-20334] Introduce module YAML format version 3.0 and function endpoint templating

tzulitai:
This PR adds a module YAML format version, 3.0, which supports function endpoint templating.

With 3.0, users define function endpoints like so:
```
endpoints:
   
  ## An endpoint url template for all function types under namespace "com.foo.bar"
  - endpoint:
      meta:
          kind: http
      spec:
          typename:
              namespace: com.foo.bar
          urlPathTemplate: http://bar.foo.com/{typename.function}
          maxNumBatchRequests: 1000
          timeouts:
              call: 5mins
              read: 10seconds

  ## An endpoint url specifically for the function type "(com.foo.bar, hello-world)"
  - endpoint:
      meta:
         kind: http
      spec:
         typename:
           namespace: com.foo.bar
           function: hello-world
         urlPathTemplate: http://bar.foo.com/some/specific/path
```

Endpoint URL templates that have been defined for specific types (fully specifying both namespace and function name) will have precedence other ones that have been widely defined for a namespace.

## Implementation notes

The main changes required to enable function templating (besides implementing the JSON parser for the new format) was the following:

- fd361ca In the Java SDK, allow binding `StatefulFunctionProvider` per namespace. Providers bound for specific function types have higher precedence.

- 42cdb52 Logic of resolving the `StatefulFunctionProvider` to use has been updated in the `PredefinedFunctionLoader`.

- 326b56a to 6c2ae2f covers the main work around introducing the new YAML format. Function endpoint specifications are represented by `FunctionEndpointSpec`, with an HTTP implementation `HttpFunctionEndpointSpec`. A new type of function provider `TemplatedHttpFunctionProvider` is used for instantiating `RequestReplyFunction`s based on the parsed `HttpFunctionEndpointSpec`.

- 45f9efb to 1d83341 updates all examples and E2E tests to use module YAML format version 3.0.

## Testing this change

The E2E `ExactlyOnceWithRemoteFnE2E` has been updated to use function endpoint templating, so a CI pass would verify the feature end-to-end.

### DIFF
--- a/statefun-e2e-tests/statefun-exactly-once-remote-e2e/src/test/resources/remote-module/module.yaml
+++ b/statefun-e2e-tests/statefun-exactly-once-remote-e2e/src/test/resources/remote-module/module.yaml
@@ -13,26 +13,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-version: "1.0"
+version: "3.0"
 
 module:
   meta:
     type: remote
   spec:
-    functions:
-      - function:
+    endpoints:
+      - endpoint:
           meta:
             kind: http
-            type: org.apache.flink.statefun.e2e.remote/counter
           spec:
-            endpoint: http://remote-function:8000/service
-            maxNumBatchRequests: 10000
-      - function:
-          meta:
-            kind: http
-            type: org.apache.flink.statefun.e2e.remote/forward-function
-          spec:
-            endpoint: http://remote-function:8000/service
+            typename:
+              namespace: org.apache.flink.statefun.e2e.remote
+            urlPathTemplate: http://remote-function:8000/service
             maxNumBatchRequests: 10000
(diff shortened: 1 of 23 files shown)


igalshilman (commented):
@tzulitai, thanks for opening this PR. The user facing changes look good to me, and I'm excited about this new feature!

tzulitai:
Thanks for the comments @igalshilman.
I addressed your comments with one minor difference:
instead of calling it `FunctionTypeNamespaceSelector`, I found that `FunctionTypeNamespaceMatcher` made more sense.

Basically, it conveys that if a message's target function matches the provided matcher, then that function provider will be used.
The same for a potential `FunctionTypeNamespacePrefixMatcher` in the future.

PR #184, opened 2020-12-04
[FLINK-20335] Remove support for module YAML versions 1.0 / 2.0

tzulitai:
This reason for this removal is to avoid user confusion of having 2 different ways to declare state specs, either eagerly using the old version module YAML formats, or via the SDKs.

After this removal, users will only be able to declare state specs in SDKs.

## Change log

- 56a0976 to 9cb8d26: The gist of the change - on startup, if a module is using versions 1.0 / 2.0 format, an error is thrown. Also, remove the JSON parser for the `functions` entity used in older versions, as well as all tests against the old format.

- 155ae39 to e0548c0: Remove all outdated internal representations such as `StateSpec` / `HttpFunctionSpec` etc. Also fully replace `HttpFunctionProvider` with the more powerful `TemplatedHttpFunctionProvider` (renamed).

- 09146ad: Now that we no longer have eager state declarations, the constructor in `PersistedRemoteFunctionValues` that accepts eager state specs is also officially removed.

- d23e0cd: Make `FunctionEndpointSpec` serializable. This is in preparation for reflecting all the recent changes around the dynamic functions epic in the DataStream interop SDK.

- 4cef4f1: Update the DataStream interop SDK to use the new internal abstractions.


igalshilman (approved):
Thanks for removing the old module formats @tzulitai,
The changes look good to me, I've built locally and ran the e2e tests.

PR #186, opened 2020-12-11
[FLINK-20574] [core] Block further inputs until first invocation completes

tzulitai:
This PR adds throttling to the `RequestReplyFunction`, to mitigate heavily loading remote functions on startup or restores where there is a large backlog of events to process through.

The change is rather simple: on receiving the first input for a given function type, dispatch the invocation as usual but block further inputs (by calling `context.awaitResponseIfNotBootstrapped`) until the request is complete.

I would suggest to only merge this after we have a version of the `SmokeE2E` that runs with remote functions.


igalshilman:
How about we'll close this PR for now?

tzulitai:
Yes makes sense!

PR #187, opened 2020-12-17
[FLINK-20636] Validate that unaligned checkpoints is not enabled

tzulitai:
Due to how StateFun has feedback loops, with unaligned checkpoints a function dispatcher operator may receive a feedback checkpoint barrier (from other parallel subtasks that process the checkpoint barrier first) before it receives its own checkpoint barrier. This currently isn't handled.

We need to further investigate how to properly support unaligned checkpointing in StateFun.
For the time being, we should strictly require aligned checkpointing for StateFun apps.


igalshilman (approved):
LGTM @tzulitai!

PR #189, opened 2020-12-21
[FLINK-20689] Upgrade to Flink 1.11.3

tzulitai:
This PR upgrades Flink version to 1.11.3.

Moreover, the `isUsingCustomRawKeyedState` flag is being set on the `FunctionGroupOperator`. Setting this flag will now (finally) enable us to safely restore from savepoints from versions earlier than 2.2.1, and containing feedback elements.

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -32,7 +32,7 @@ version: "2.3-SNAPSHOT"
 # release this should be the same as the regular version
 version_title: "2.3-SNAPSHOT"
 # The Flink version supported by this version of Stateful Functions
-flink_version: "1.11.1"
+flink_version: "1.11.3"
 # Branch on Github for this version
 github_branch: "master"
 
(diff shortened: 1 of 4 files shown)


igalshilman (approved):
Thanks @tzulitai , LGTM.
One comment tho, grepping the directory for 1.11.* show that following comments in 
config.sh and taskmanager.sh:

```
# Backported due to FLINK-18639
# This file can be removed when upgrading
# to Flink 1.11.2 or 1.12
```
I think that these were added by @sjwiesman, and it would be nice to remove them, if they are not needed anymore.

sjwiesman:
Those files were added by me, and yes we should drop them.

sjwiesman:
Also (and more importantly) remove those files from apache/flink-statefun-docker

tzulitai:
Thanks, will remove those no longer needed files.

PR #191, opened 2020-12-23
[FLINK-20742] Skip deployment of Maven artifacts for examples

tzulitai:
This PR makes us no longer deploy / release Maven artifacts for the examples, starting from the next release.
We never expect users to try out the examples via Maven artifacts.

## Brief change log

- 498bebc Configure the `maven-deploy-plugin` to skip deployment for `statefun-examples`
- 35de6fb Is a nice cleanup we can now do to the packaging setup of the `statefun-ridesharing-example-simulator`. Before, since we were publishing Maven artifacts, we had to use the Maven shade plugin for packaging because `spring-boot-maven-plugin` doesn't support adding custom transformers which was required for merging NOTICE and license files for ASF legal
purposes. Since we now no longer release these artifacts, we can revert back to use the simple `spring-boot-maven-plugin` for the packaging the simulator.

### DIFF
--- a/statefun-examples/pom.xml
+++ b/statefun-examples/pom.xml
@@ -40,5 +40,16 @@ under the License.
         <module>statefun-flink-datastream-example</module>
     </modules>
 
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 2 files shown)


igalshilman (approved):
Thanks @tzulitai, yes let's not deploy the examples. 👍

PR #195, opened 2021-01-29
[FLINK-21171] Wire in TypedValue throughout the runtime as state values and message payloads

tzulitai:
The end result of this PR is as follows:
- The `RequestReplyFunction` now fully works against `TypedValue`s, and not Protobuf `Any`s. This means that messages sent from functions to other functions as well as egresses, and state values, are now all `TypedValue`s.
- Messages sent from ingresses to functions via the text-based auto-routable Kafka / Kinesis ingresses, are also `TypedValue`s now.
- The Python SDK has been modified minimally to work with the new type system. This means that, at the API surface, nothing is changed and users still work with `Any`. To work with the new `TypedValue`s, at the very boundaries of the SDK we convert `TypedValuie` to / from `Any`. This is a temporary solution to minimize the changes in this PR, and will go away once we rework the Python SDK to work properly with the new type system.

---

# Verifying this change

- E2E tests are passing ✅ 
- Some new unit tests for the `PersistedRemoteFunctionValues` are added for coverage on the user contract regarding mutating state values with a mismatching typename.
- Manually running the Python greeter example should also work properly.


igalshilman (approved):
Thanks @tzulitai LGTM

PR #198, opened 2021-02-04
[FLINK-21276] [legal] Mention Protobuf BSD license in statefun-sdk-java

tzulitai:
Making sure ASF legal-wise we are good with the new `statefun-java-sdk` artifact.
Since we are bundling and shading Protobuf, that should be acknowledged in the NOTICE file and also the BSD license should be bundled in the artifact.


igalshilman (approved):
Thanks @tzulitai , LGTM

PR #199, opened 2021-02-08
[FLINK-21324] fix assertions for messages function sends to itself

f1xmAn:
Here is the bug this PR fixes https://issues.apache.org/jira/browse/FLINK-21324


tzulitai:
I do agree that the contract of `FunctionTestHarness.invoke()` should include messages sent to the same function instance, as part of the returned responses map.

@sjwiesman what's your take here? Would like to confirm before I merge this.

sjwiesman:
+1 This was a bug on my part. LGTM.

sjwiesman (approved):

PR #200, opened 2021-02-19
[FLINK-21422][docs] Migrate docs to Hugo 

sjwiesman:
This PR migrates statefun documentation to Hugo from Jekyll. 

While we will be doing a major revamp of the content of the documentation prior to 3.0, this PR is solely an infrastructure change. We can make follow up updates to the style and substance going forward. 

8ad3747   Contains the actual migration. I would not look at this diff. Instead, I encourage you to build the docs locally and click around. There are instructions in `docs/README.md`. Additionally, all the layout and hugo specific files were copied directly from apache/flink and have already been reviewed. 

3dc1559 Contains some changes to build and release scripts. Please do review this diff :) 

### Build Bot

After merging this change we will need to update the build bot. Unfortunately, I do not have access as it is PMC only. I have fully scripted the build bot steps. A PMC member simply needs to update it to run `./tools/ci/docs.sh` to build and deploy. 

**IMPORTANT** Jekyll and Hugo both use a directory called docs/content for different purposes. Jekyll uses it to cache build artifacts and as such it has been part of flinks .gitignore. This means the Jekyll content will not be removed when you check out this branch. Please manually delete these files or the site will fail to build. I strongly encourage you to delete docs/content before checking out this branch.


sjwiesman:
rebased on master

tools/ci/docs.sh
```diff
@@ -17,29 +17,22 @@
...
-if [ ! -f ${PROJECT_ROOT}/docs/build_docs.sh ]; then
-    echo "Project root path ${PROJECT_ROOT} is not valid; script may be in the wrong directory."
-    exit 1
+if ! curl --fail -OL $HUGO_REPO ; then 
+	echo "Failed to download Hugo binary"
+	exit 1
 fi
 
-CACHE_DIR=$HOME/gem_cache ${PROJECT_ROOT}/docs/build_docs.sh -p &
+tar -zxvf $HUGO_ARTIFACT
```
tzulitai, line 28: I think it would be safer to extract the hugo artifact in a separate directory, just in case it somehow pollutes the docs dir. I don't think that's the case right now, but would be good to do.

tzulitai (commented):
Just one minor comment. WDYT?

tools/ci/docs.sh
sjwiesman: I've updated it to only extract the specific binary file we need.

sjwiesman:
@tzulitai thanks for the review. I've update the build bot script. I will leave this one for your to merge since it requires the build bot update. Please let me know if you have any questions.

tzulitai:
Looks good to merge! Will merge this and update the buildbot now.

PR #201, opened 2021-02-23
[FLINK-21459] [sdks] Implement remote Java SDK

tzulitai:
Add initial implementation of the revisioned remote Java SDK, based on the new remote invocation protocol.

The goal of this implementation is to serve as a base minimalistic library, and more complex SDKs such as a Spring Boot Java SDK can be built upon this.

### DIFF
--- a/statefun-flink/statefun-flink-common/src/main/java/org/apache/flink/statefun/flink/common/types/TypedValueUtil.java
+++ b/statefun-flink/statefun-flink-common/src/main/java/org/apache/flink/statefun/flink/common/types/TypedValueUtil.java
@@ -36,6 +36,7 @@ public final class TypedValueUtil {
   public static TypedValue packProtobufMessage(Message protobufMessage) {
     return TypedValue.newBuilder()
         .setTypename(protobufTypeUrl(protobufMessage.getDescriptorForType()))
+        .setHasValue(true)
         .setValue(protobufMessage.toByteString())
         .build();
   }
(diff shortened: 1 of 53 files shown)


sjwiesman:
🔥 Very excited to see this. I did not do a code review but I spent today building a demo to get my head in the right space for the new docs and I have two nits of things I found surprising as a user. I'll leave it up to you two if you think my suggestions make sense. 

1) Replace `SimpleType#simpleTypeFrom` and `SimpleType#simpleImmutableTypeFrom` with a builder. Most everything else in the user API, like ValueSpec and StatefulFunctions, follows a builder pattern and I found it surprising that this was different. I expected to find something like: 

```java
SimpleType
   .named(MyTypeName)
   .withSerializer(mySerializer, myDeserializer)
   .isImmutable();
```

2) `RequestReplyHandler#handle` taking and returning a `Slice` threw me for a loop. Especially because it is not obvious how to turn a byte[] into a `Slice`. I had to find the helper class by digging through the code. Maybe this one is resolved with some additional Javadoc but it's also not clear to me why `Slice` is exposed to the user at all.

tzulitai:
The CI was failing due to E2Es (incompatibility between the Testcontainer usage of the Docker Java API + the Docker version used by Github Actions). I've disabled the E2E tests for the time being, and rebasing this PR on top of that.

tzulitai:
@sjwiesman `Slice` is exposed through the `TypeSerializer` as well, and exposing it would actually allow us to avoid excessive byte array copies end-to-end. @igalshilman would know best about this.

For building simple types, that's a good point. I would prefer not to touch the PR anymore though, and we can maybe address that as a follow-up. There's quite a few follow-up tasks pending on the Java SDK being merged.

tzulitai:
CI is passing, merging this.
I'll return back to fixing the CI for E2E tests right after this.

igalshilman:
@sjwiesman thanks for the suggestion! Yes a builder here will read nicer.
I'll followup with that.
Regarding the `Slice`, the main motivation was to enable efficient interopabilty with the embedding HTTP server.
(each framework

PR #202, opened 2021-02-24
[FLINK-21491] Properly shade Protobuf in Java SDK

tzulitai:
This PR contains mainly of 2 important changes:
- Create our own modules that shades Protobuf + declare the `protobuf-java` dependency (which is required only by the protoc maven plugin) to be provided, so that it doesn't get pulled in as a transitive dependency by users.
- Let the Java SDK depend on this new module, instead of `protobuf-java` directly.

---

# Verifying this change

- Running `mvn dependency:tree` under `statefun-sdk-java` should reveal that `com.google.protobuf:protobuf-java` is no longer a compile dependency.
- After building the project + reloading Maven in the IDE, there should be no warnings or errors in the IDE, meaning that all references in the Java SDK to Protobuf classes and protocol message Java classes are pointing to already relocated sources.


igalshilman (approved):
LGTM

PR #203, opened 2021-02-25
[FLINK-21498] Avoid copying when converting byte[] to ByteString

tzulitai:
There's a few places where we can be more efficient by avoiding byte array copying and we know that it is safe to do so (because we won't be mutating the byte array):

- The message payload serializers
- In `PersistedRemoteFunctionValues`, when attaching state bytes to a ToFunction
- In deserializers used by auto-routable Kafka / Kinesis ingresses.

---

## Verifying the change

E2E tests pass

### DIFF
--- a/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/message/MessagePayloadSerializerKryo.java
+++ b/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/message/MessagePayloadSerializerKryo.java
@@ -18,6 +18,7 @@
 package org.apache.flink.statefun.flink.core.message;
 
 import com.google.protobuf.ByteString;
+import com.google.protobuf.MoreByteStrings;
 import java.io.IOException;
 import javax.annotation.Nonnull;
 import org.apache.flink.api.common.ExecutionConfig;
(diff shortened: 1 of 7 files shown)


igalshilman (approved):
LGTM

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/message/MessagePayloadSerializerKryo.java
```diff
@@ -40,8 +41,7 @@ public Payload serialize(@Nonnull Object payloadObject) {
     } catch (IOException e) {
       throw new IllegalStateException(e);
     }
-    // TODO: avoid copying, consider adding a zero-copy ByteString.
-    ByteString serializedBytes = ByteString.copyFrom(target.getSharedBuffer(), 0, target.length());
+    ByteString serializedBytes = MoreByteStrings.wrap(target.getSharedBuffer(), 0, target.length());
```
igalshilman, line 44: @tzulitai - just a reminder: please undo this change before you merge.

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/message/MessagePayloadSerializerRaw.java
```diff
@@ -33,7 +34,7 @@ public Object deserialize(@Nonnull ClassLoader targetClassLoader, @Nonnull Paylo
   @Override
   public Payload serialize(@Nonnull Object what) {
     byte[] bytes = (byte[]) what;
-    ByteString bs = ByteString.copyFrom(bytes);
+    ByteString bs = MoreByteStrings.wrap(bytes);
```
igalshilman, line 37: same here.

PR #204, opened 2021-03-01
[FLINK-21530] Precompute TypeName's canonical representation

tzulitai:
There's room for improvement in `PersistedRemoteFunctionValues`, where we currently concatenate strings to build the typename string for each state value we attach to a `ToFunction` message.

This extra work can be easily avoided by precomputing the canonical typename string, since `TypeName`'s are immutable.


igalshilman (approved):
LGTM +1

PR #205, opened 2021-03-02
[hotfix] [java] Set has_value appropriately for TypedValues

tzulitai:
Did a grep for `TypedValue.newBuilder` and `TypedValue.Builder` to see where we are constructing `TypedValue`s, and make sure we appropriately set the `has_value` flag (by default, the flag is `false`).

It doesn't really matter for invocation arguments for now, since we don't do any explicit checks on the flag for arguments.
This is important only for keeping the semantics of the flag consistent across invocation argument values and state values.


igalshilman (approved):
LGTM!

PR #206, opened 2021-03-03
[FLINK-21577] [java] Fix instantiation error with SimpleType.simpleTypeFrom()

tzulitai:
This was failing with an `Collection is empty` when attempting to copy the enum set of `TypeCharacteristic`s.

I think the only reason we need to do a copy, is if users instantiate a `SimpleType` directly with the constructor (and not the `simpleTypeFrom`, `simpleImmutableTypeFrom` factory methods).
IMO, we probably can expect users to just use the factory methods, and not the constructor.
If they need to do anything more complex, they should just create a `Type` subclass directly.

Therefore, this PR makes the constructor private to enforce instantiation a `SimpleType` via the factory methods.
This would allow the constructor to be simpler and remove the `EnumSet.copyOf` operation.

### DIFF
--- a/statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/types/SimpleType.java
+++ b/statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/types/SimpleType.java
@@ -47,14 +47,14 @@ public final class SimpleType<T> implements Type<T> {
   private final TypeSerializer<T> serializer;
   private final Set<TypeCharacteristics> typeCharacteristics;
 
-  public SimpleType(
+  private SimpleType(
       TypeName typeName,
       Fn<T, byte[]> serialize,
       Fn<byte[], T> deserialize,
       Set<TypeCharacteristics> typeCharacteristics) {
     this.typeName = Objects.requireNonNull(typeName);
     this.serializer = new Serializer<>(serialize, deserialize);
-    this.typeCharacteristics = Collections.unmodifiableSet(EnumSet.copyOf(typeCharacteristics));
+    this.typeCharacteristics = Collections.unmodifiableSet(typeCharacteristics);
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


statefun-sdk-java/src/test/java/org/apache/flink/statefun/sdk/java/types/SimpleTypeTest.java
```diff
@@ -0,0 +1,63 @@
...
+import static org.junit.Assert.assertThat;
+
+import org.apache.flink.statefun.sdk.java.TypeName;
+import org.apache.flink.statefun.sdk.java.slice.Slice;
+import org.junit.Test;
+
+public class SimpleTypeTest {
+
+  @Test
+  public void mutableType() {
```
igalshilman, line 32: Minor comment here, regarding `String::getBytes`: I know that this is irrelevant for this test in particular, 
but since this is part of the SDK test, I wouldn't want users to get the wrong impression that this is how we recommend dealing with String (folks have the habit to copy paste :-P) 
Can you replace the method references with a version that accepts a UTF8 charset?

statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/types/SimpleType.java
```diff
@@ -47,14 +47,14 @@
...
 
-  public SimpleType(
+  private SimpleType(
       TypeName typeName,
       Fn<T, byte[]> serialize,
       Fn<byte[], T> deserialize,
       Set<TypeCharacteristics> typeCharacteristics) {
     this.typeName = Objects.requireNonNull(typeName);
     this.serializer = new Serializer<>(serialize, deserialize);
-    this.typeCharacteristics = Collections.unmodifiableSet(EnumSet.copyOf(typeCharacteristics));
```
igalshilman, line 57: 👍

igalshilman (approved):
Thanks @tzulitai , good catch!

tzulitai:
Thanks for the review @igalshilman!
Will address your comment and merge this.

PR #215, opened 2021-03-16
[FLINK-21826] Migrate the shopping-cart-example to the new SDK.

afedulov:
This PR migrates the shopping-cart-example to the new SDK.
I've added `module.yaml` for verifying a potential configuration.
As the next step it would be nice to make this example runnable and to check that the JSON un/marshalling and the wiring work as expected. There is also an open question regarding multiplexing related JSON messages via the same topic (see `module.yaml`).


afedulov:
I am closing this one because this PR is now open against the `flink-statefun-playground` repository:
https://github.com/apache/flink-statefun-playground/pull/2

PR #223, opened 2021-04-12
[FLINK-22242][python] Improve missing attribute error on GeneratedAddressedScopedStorage

sjwiesman:
```python
@functions.bind('example/func', [ValueSpec('state', IntType)]
def my_func(ctx, msg): 
     ctx.storage.sate
```

Previous Error: 
> AttributeError: No attribute sate in type GeneratedAddressedScopedStorage

New Error: 

> AttributeError: 'sate' is not a registered ValueSpec for the function 'example/func'

### DIFF
--- a/statefun-sdk-python/statefun/request_reply_v3.py
+++ b/statefun-sdk-python/statefun/request_reply_v3.py
@@ -209,7 +209,7 @@ class RequestReplyHandler(object):
         if not target_fn:
             raise ValueError(f"Unable to find a function of type {sdk_address.typename}")
         # resolve state
-        res = resolve(target_fn.storage_spec, pb_to_function.invocation.state)
+        res = resolve(target_fn.storage_spec, sdk_address.typename, pb_to_function.invocation.state)
         if res.missing_specs:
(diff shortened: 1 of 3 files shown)


igalshilman (changes requested):
Hey @sjwiesman! I like the idea of improving the error message! and I think that it will be even more effective to do it at the `RequestReplyHandler`.

Essentially any error that might happen during the invocation of a function needs to be augmented with the currently executing address, function type, etc'.

We have it in the embedded SDK [here](https://github.com/apache/flink-statefun/blob/master/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/functions/StatefulFunction.java#L44,%20L50) , and definitely we need it in the remote ones as well.

Specifically to the change that you are proposing here, I'd be very careful about adding additional properties to this generated class because it increases the chances of collision of user supplied `ValueSpec`.

igalshilman (approved):
Hi @sjwiesman 
I've looked into it again, and verified that adding `getattr` doesn't change the characteristics of the storage class.
So I'd say lets merge that, it is 100% an improvement in usability.

PR #226, opened 2021-04-17
Update java.md

birbilis:
Changed Python mention to Java, probably left-over from python doc


birbilis:
not sure what the second change at the "keep learning" shown in the diff is about, used GitHub's online editor for the patch

tzulitai:
Merging, thanks @birbilis

PR #230, opened 2021-04-26
[FLINK-22468] Fix Java 11 build by adding dependency to javax.annotations

tzulitai:
This enables building StateFun with Java 11, since javax.annotations was removed with JDK 11.


knaufk (approved):
LGTM. Successfully `mvn clean install` with OpenJDK 11.

tzulitai:
@knaufk thanks for the review and trying this out.

I'm not merging this just yet because we should make this a profile that is enabled specifically for Java 11 builds.
Or, the other way around, where we build with Java 11 by default and remove the javax dependency when building with Java 8.

PR #231, opened 2021-05-03
 [FLINK-22554] Add topic subscription to KafkaIngress 

sjwiesman:
Update the Kafka ingress to support Flinks Kafka topic discovery. This removes a common source of downtime from the statefun cluster, configuring new topics.  Users can now configure their ingress to use a topic pattern and the source will consume from all topics that match the regex and auto-discover new topics every `discoveryInterval`. 

```yaml
 version: "3.0"
 
 module:
   meta:
     type: remote
 spec:
   ingresses:
   - ingress:
       meta:
         type: io.statefun.kafka/ingress
         id: com.example/users
       spec:
         address: kafka-broker:9092
         consumerGroupId: my-consumer-group
         startupPosition:
           type: earliest
         topicPattern:
            pattern: topic-*
            discoveryInterval: 30s
            valueType: com.example/User
            targets:
              - com.example.fns/greeter
```


tzulitai:
I took a pass over the implementation, and while this would work I'm wondering about how useful this feature would be with the current way of how things work.

With this PR, we are enabling routing topics that match a specific pattern to a static set of functions.
I believe the value of this feature really shines when, we add a new function to the topology, and would like it to receive events from a new Kafka topic as well. This is not enabled by the proposal in this PR.

If there is already a use case out there that could make use of this PR, I'm fine to merge this, but if not, I'd prefer to take a step back and think if we can address this by looking at a more macro scope of the dynamic functions story.

sjwiesman:
@tzulitai that makes a lot of sense to me. I’ll close this for now

PR #234, opened 2021-05-18
[FLINK-22552] Update Flink dependency to 1.13.0

tzulitai:
This PR rebases StateFun to be compatible with Flink 1.13.x.
While the PR upgrades to 1.13.0, ultimately, we want to at least update to 1.13.1. We may postpone the merge of this PR until 1.13.1 is out, or already merge this so that once 1.13.1 is out it'll be just a simple version bump.


igalshilman (approved):
LGTM @tzulitai !

PR #236, opened 2021-05-25
[FLINK-22771] Add TestContext to Java SDK

knaufk:
There is the fundamental question of whether this should go into it's only testing module. Feels like overkill to me, though.

### DIFF
--- a/statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/testing/Envelope.java
+++ b/statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/testing/Envelope.java
@@ -1,0 +1,68 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.flink.statefun.sdk.java.testing;
+
+import java.time.Duration;
+import java.util.Objects;
+import org.apache.flink.statefun.sdk.java.message.Message;
+
+/**
+ * A utility class that wraps a {@link Message} and the {@link Duration} after which it was sent by
+ * a {@link org.apache.flink.statefun.sdk.java.StatefulFunction}. It is used by the {@link
+ * TestContext}.
+ */
+public class Envelope {
+
+  private final Duration duration;
+  private final Message message;
+
+  public Envelope(Duration duration, Message message) {
+    this.duration = duration;
+    this.message = message;
+  }
+
+  public Duration getDuration() {
+    return duration;
+  }
+
+  public Message getMessage() {
+    return message;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+    Envelope that = (Envelope) o;
+    return Objects.equals(duration, that.duration) && Objects.equals(message, that.message);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(duration, message);
+  }
+
+  @Override
+  public String toString() {
+    final StringBuilder sb = new StringBuilder("SentMessage{");
+    sb.append("duration=").append(duration);
+    sb.append(", message=").append(message);
+    sb.append('}');
+    return sb.toString();
+  }
+}
--- a/statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/testing/TestAddressScopedStorage.java
+++ b/statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/testing/TestAddressScopedStorage.java
@@ -1,0 +1,44 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.flink.statefun.sdk.java.testing;
+
+import java.util.HashMap;
+import java.util.Optional;
+import org.apache.flink.statefun.sdk.java.AddressScopedStorage;
+import org.apache.flink.statefun.sdk.java.ValueSpec;
+
+class TestAddressScopedStorage implements AddressScopedStorage {
+
+  private HashMap<ValueSpec, Object> storage = new HashMap<>();
+
+  @Override
+  public <T> Optional<T> get(ValueSpec<T> spec) {
+
+    return (Optional<T>) Optional.ofNullable(storage.get(spec));
+  }
+
+  @Override
+  public <T> void set(ValueSpec<T> spec, T value) {
+    storage.put(spec, value);
+  }
+
+  @Override
+  public <T> void remove(ValueSpec<T> spec) {
+    storage.remove(spec);
+  }
+}
--- a/statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/testing/TestContext.java
+++ b/statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/testing/TestContext.java
@@ -1,0 +1,111 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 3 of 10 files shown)


sjwiesman:
I think this is fine where it is. My threshold for moving it to a dedicated module would be if we add custom Matchers, or pull in any sort of testing dependency.

statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/testing/Envelope.java
```diff
@@ -0,0 +1,68 @@
...
+
+  private final Duration duration;
+  private final Message message;
+
+  public Envelope(Duration duration, Message message) {
+    this.duration = duration;
+    this.message = message;
+  }
+
+  public Duration getDuration() {
```
sjwiesman, line 39: Can you add a `@Nullable` annotation

sjwiesman (commented):
This looks good to me, can you also add docs? Otherwise, I fear no one will ever find this.

statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/testing/Envelope.java
knaufk: Which one? It does not seem to be on the classpath so far.

knaufk:
> This looks good to me, can you also add docs? Otherwise, I fear no one will ever find this.

Thank you. I've added docs.

statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/testing/TestContext.java
```diff
@@ -0,0 +1,111 @@
...
+   * Context#send(Message)} or {@link Context#sendAfter(Duration, Message)}.
+   *
+   * <p>Messages are wrapped in an {@link Envelope} that contains the message itself and the
+   * duration after which the message was sent. The Duration is {@link Duration#ZERO} for messages
+   * sent via {@link Context#send(Message)}.
+   *
+   * @return the list of sent messages wrapped in {@link Envelope}s
+   */
+  public List<Envelope> getSentMessages() {
+    return sentMessages;
```
igalshilman, line 99: perhaps you'd like to make a copy here or wrap in a `Collections.unmodifiableList`

statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/testing/Envelope.java
```diff
@@ -0,0 +1,68 @@
...
+import java.time.Duration;
+import java.util.Objects;
+import org.apache.flink.statefun.sdk.java.message.Message;
+
+/**
+ * A utility class that wraps a {@link Message} and the {@link Duration} after which it was sent by
+ * a {@link org.apache.flink.statefun.sdk.java.StatefulFunction}. It is used by the {@link
+ * TestContext}.
+ */
+public class Envelope {
```
igalshilman, line 29: I like the idea of an envelope, but I think that this class is a bit confusing as it, requires a delay, and generally a bit tricky to extend in the future (make it capture a `context` method).
How about we'll just split it to `SentEgressMessage` , `SentMessage`, and `SentDelayedMessage` ?

They don't have to implement a common interface) but just 3 separate classes to capture the result of 3 different `context` invocations.

statefun-sdk-java/src/test/java/org/apache/flink/statefun/sdk/java/testing/TestContextIntegrationTest.java
```diff
@@ -0,0 +1,114 @@
...
+    static final TypeName ANOTHER_TYPE = TypeName.typeNameFromString("com.example.fns/another-fn");
+
+    static final TypeName SOME_EGRESS = TypeName.typeNameFromString("com.example.fns/another-fn");
+
+    static final ValueSpec<Integer> NUM_INVOCATIONS = ValueSpec.named("seen").withIntType();
+
+    @Override
+    public CompletableFuture<Void> apply(Context context, Message message) throws Throwable {
+
+      String name = message.asUtf8String();
```
igalshilman, line 50: This seems to be unused

statefun-sdk-java/src/test/java/org/apache/flink/statefun/sdk/java/testing/TestContextIntegrationTest.java
```diff
@@ -0,0 +1,114 @@
...
+              .build());
+
+      context.sendAfter(Duration.ofMillis(1000), messageToSomeone);
+
+      return context.done();
+    }
+  }
+
+  @Test
+  public void testSimpleFunction() throws Throwable {
```
igalshilman, line 74: nice tests!

sjwiesman:
Docs look good. Maybe we can use Optional instead of an annotation to avoid additional dependencies, your call. Otherwise +1

statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/testing/TestAddressScopedStorage.java
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+package org.apache.flink.statefun.sdk.java.testing;
+
+import java.util.HashMap;
+import java.util.Optional;
+import org.apache.flink.statefun.sdk.java.AddressScopedStorage;
+import org.apache.flink.statefun.sdk.java.ValueSpec;
+
+class TestAddressScopedStorage implements AddressScopedStorage {
```
igalshilman, line 25: It is definitely useful to keep the objects like that in their original form as a reference, but I'm wondering if we should go trough serde here. Otherwise the test might be a little misleading. 

While it is true that testing serialization should happen else-where, but I'm just wondering how many people would follow these guidelines.

igalshilman (commented):
Hey @knaufk that's for working on this! this is a useful addition for sure :-)

* Regarding rather or not that should be a separate module - I agree with @sjwiesman, lets leave it here for now.

* I think that the overloaded `Envelope` is a bit problematic and it is better to replace it with explicit classes `SentMessage`, `SentDelayedMessage`, etc`.

* The storage class is nice but it is missing the serialization round trip. Although this is a bit out of scope, I think that it should be here, as it can help folks detect problems earlier.

statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/testing/TestContext.java
```diff
@@ -0,0 +1,111 @@
...
+import org.apache.flink.statefun.sdk.java.Context;
+import org.apache.flink.statefun.sdk.java.message.EgressMessage;
+import org.apache.flink.statefun.sdk.java.message.Message;
+
+/**
+ * An implementation of {@link Context} to to make it easier to test {@link
+ * org.apache.flink.statefun.sdk.java.StatefulFunction}s in isolation. It can be instantiated with
+ * the address of the function under test and optionally the address of the caller.
+ */
+public class TestContext implements Context {
```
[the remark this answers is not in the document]
knaufk: Done.

statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/testing/TestContext.java
```diff
@@ -0,0 +1,111 @@
...
+  private List<Envelope> sentMessages = new ArrayList<>();
+  private List<EgressMessage> sentEgressMessages = new ArrayList<>();
+
+  private TestContext(Address self, Optional<Address> caller) {
+    this.self = self;
+    this.caller = caller;
+    this.storage = new TestAddressScopedStorage();
+  }
+
+  public TestContext(Address self) {
```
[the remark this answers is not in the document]
knaufk: Done.

statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/testing/TestContext.java
```diff
@@ -0,0 +1,111 @@
...
+    this(self, Optional.empty());
+  }
+
+  public TestContext(Address self, Address caller) {
+    this(self, Optional.of(caller));
```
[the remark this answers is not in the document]
knaufk: Done

statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/testing/TestContext.java
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  @Override
+  public Optional<Address> caller() {
+    return caller;
+  }
+
+  @Override
+  public void send(Message message) {
+    sentMessages.add(new Envelope(Duration.ofMillis(0), message));
```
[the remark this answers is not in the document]
knaufk: Done

statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/testing/TestContext.java
knaufk: Done.

statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/testing/TestContext.java
```diff
@@ -0,0 +1,111 @@
...
+ * org.apache.flink.statefun.sdk.java.StatefulFunction}s in isolation. It can be instantiated with
+ * the address of the function under test and optionally the address of the caller.
+ */
+public class TestContext implements Context {
+
+  private final TestAddressScopedStorage storage;
+  private Address self;
+  private Optional<Address> caller;
+
+  private List<Envelope> sentMessages = new ArrayList<>();
```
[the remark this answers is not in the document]
knaufk: Done.

statefun-sdk-java/src/test/java/org/apache/flink/statefun/sdk/java/testing/TestContextIntegrationTest.java
knaufk: Removed.

statefun-sdk-java/src/test/java/org/apache/flink/statefun/sdk/java/testing/TestContextIntegrationTest.java
knaufk: Thanks.

statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/testing/TestAddressScopedStorage.java
knaufk: Do you suggest to use `ConcurrentAddressScopedStorage` instead of the TestAdressScopedStorage?

statefun-sdk-java/src/main/java/org/apache/flink/statefun/sdk/java/testing/Envelope.java
knaufk: Done.

tzulitai:
@igalshilman @knaufk what's the status here? Should this be merged?

igalshilman:
@tzulitai I will merge this after few minor fix-ups.

f1xmAn:
Guys, I'm considering building a BDD-style assertions library for TestContext. Here is a possible usage example:
```java
given(context).havingValues(
    SPEC1, value1,
    SPEC2, value2
)

FunctionUnderTest function = new FunctionUnderTest();
Message message = MessageBuilder
    .forAddress(self)
    .withValue("This is a message")
    .build();
function.apply(context, message);

then(context).gotMessage(expectedMessage);
then(context).hasValue(SPEC1, expectedValue1);
```

Do you think it might be useful? Or perhaps you already built something similar?

PR #237, opened 2021-06-02
[FLINK-22529] [kinese] Allow Flink's ConsumerConfigConstants and flexibility in providing AWS region and credentials

tzulitai:
This PR makes two major changes to the Kinesis ingress:

1. Configurations in Flink Kinesis connector's `ConsumerConfigConstants` are now applicable, by directly providing them in the properties.
2. AWS region and credentials can now be provided purely from the properties. If users do choose to specify something using the named methods, those values have higher precedence. If nothing was specified in either the properties or via named methods, then the default is to fallback to credentials provided in the environment (of AWS instances, for example).


igalshilman (approved):
Thanks @tzulitai, I have a minor comment about a JavaDoc, other than that LGTM.

PR #239, opened 2021-06-09
Turn smoke e2e Fn into a RemoteFunction

evans-ye:


evans-ye:
Sorry. This is still a private draft . Shouldn't send to the public repository.

PR #243, opened 2021-07-12
[FLINK-23293] [core] Support extendable transports for remote invocations

tzulitai:
The goal of this PR is to allow pluggable / extendable transport clients for invoking remote functions.

There are a few major changes included:

- (1st commit) Add new SPI interfaces for binding extension objects by `TypeName`. Specifically, these new interfaces are `ExtensionModule` and `ExtensionResolver`, the former used for binding extensions to the application universe, and former used for getting access to bound extensions within the main modules, etc. `StatefulFunctionModules` or `FlinkIOModules`. 2cdddf9

- Introduces a new interface, `RequestReplyClientFactory`, that is responsible for creating the client for each individual remote function endpoint. This interface is public to users as an extension, by binding client factory instances using `ExtensionModule`s. Later on, the factory to use can be referenced using the `TypeName` that was used when binding the factory. a808653 84bce57 45aa1b4 c4f37da

- Furthermore, the original OkHttp client's instantiation logic has been refactored as a `RequestReplyClientFactory`. This factory is used by default if users do not specify any transport extensions. 387f7e2 84be2ba

- Last of all, we bump the module YAML format version, and rework how transport properties are specified. b6f847b af7f337

### DIFF
--- a/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/httpfn/HttpFunctionProvider.java
+++ b/statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/httpfn/HttpFunctionProvider.java
@@ -22,7 +22,9 @@ import java.util.Map;
 import java.util.Objects;
 import javax.annotation.concurrent.NotThreadSafe;
 import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.databind.node.ObjectNode;
+import org.apache.flink.statefun.flink.common.SetContextClassLoader;
 import org.apache.flink.statefun.flink.core.common.ManagingResources;
+import org.apache.flink.statefun.flink.core.reqreply.ContextSafeRequestReplyClient;
 import org.apache.flink.statefun.flink.core.reqreply.RequestReplyClient;
 import org.apache.flink.statefun.flink.core.reqreply.RequestReplyClientFactory;
 import org.apache.flink.statefun.flink.core.reqreply.RequestReplyFunction;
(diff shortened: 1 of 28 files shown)


statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/httpfn/HttpFunctionProvider.java
```diff
@@ -71,40 +68,35 @@ private HttpFunctionEndpointSpec getEndpointsSpecOrThrow(FunctionType functionTy
...
+    specificTypeEndpointSpecs
+        .values()
+        .forEach(spec -> shutdownTransportClientFactory(spec.transportClientFactory()));
+    perNamespaceEndpointSpecs
+        .values()
+        .forEach(spec -> shutdownTransportClientFactory(spec.transportClientFactory()));
+  }
+
+  private static void shutdownTransportClientFactory(RequestReplyClientFactory factory) {
+    try (SetContextClassLoader ignored = new SetContextClassLoader(factory)) {
```
tzulitai, line 98: As you can see, switching classloaders is currently a concern handled by the `HttpFunctionProvider`, and is rather error prone and fragile as we include more and more extensions in the future.

It may be worth considering moving this responsibility elsewhere, e.g. the `ExtensionResolver` should perhaps always return a classloader-safe, already decorated instance of extensions. If that's the case, we can clean up a lot of this classloader checks in the `HttpFunctionProvider`, which is essentially user code to the runtime.

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/jsonmodule/JsonModule.java
```diff
@@ -49,10 +50,17 @@ public JsonModule(JsonNode moduleSpecNode, FormatVersion formatVersion, URL modu
...
+              jsonEntity.bind(binder, getExtensionResolver(binder), moduleSpecNode, formatVersion));
     } catch (Throwable t) {
       throw new ModuleConfigurationException(
           format("Error while parsing module at %s", moduleUrl), t);
     }
   }
+
+  // TODO expose ExtensionResolver properly once we have more usages
+  private static ExtensionResolver getExtensionResolver(Binder binder) {
+    return (ExtensionResolver) binder;
```
tzulitai, line 64: Notice this TODO.

I have considered add a `getExtensionResolver()` method on the binder interface of `StatefulFunctionModule.Binder`, but wasn't sure if it the right move given how few usages we have so far.

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/jsonmodule/FunctionEndpointJsonEntity.java
```diff
@@ -146,6 +170,42 @@ private static FunctionEndpointSpec parseFunctionEndpointsSpec(
...
+            .map(TypeName::parseFrom)
+            .map(
+                extensionType ->
+                    extensionResolver.resolveExtension(
+                        extensionType, RequestReplyClientFactory.class));
+    transportClientFactory.ifPresent(endpointSpecBuilder::withTransportClientFactory);
+
+    // retain everything except "type" field, and use that directly as the transport client
+    // properties
+    transportSpecNode.remove("type");
```
tzulitai, line 188: It can be argued that remove the `type` field isn't absolutely necessary.

PR #245, opened 2021-07-28
[FLINK-23520][datastream] improve statefun <-> datastream interop

sjwiesman:
Improve Stateful Functions <-> DataStream interop to improve usability with remote functions based on best practices / lessons learned from DataStream / Table interop. In particular:

- Automatic type conversion between JVM primitives and remote type system
- Plugable types for use with complex class
- Hide internal classes like TypedValue
- Support all endpoint configuration including wildcards and URL path templating

Please see the docs and ITCase for full working examples.

Unfortunately, I had to copy the Type and TypeSerializer classes from `statefun-sdk-java` because on the DataStream side we need them working against unshaded protobuf. As I don't imagine these classes will often (if ever) change I think this is alright but I wanted to point it out.

### DIFF
--- a/statefun-flink/statefun-flink-datastream/pom.xml
+++ b/statefun-flink/statefun-flink-datastream/pom.xml
@@ -29,10 +29,21 @@ under the License.
 
     <artifactId>statefun-flink-datastream</artifactId>
 
+    <properties>
+        <protocol-messages.dir>${basedir}/target/generated-sources/protocol-messages/</protocol-messages.dir>
+        <proto-sources.dir>target/proto-sources</proto-sources.dir>
+        <protoc-jar-maven-plugin.version>3.11.1</protoc-jar-maven-plugin.version>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>org.apache.flink</groupId>
             <artifactId>statefun-sdk-embedded</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>statefun-sdk-protos</artifactId>
             <version>${project.version}</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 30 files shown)


statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/httpfn/DefaultHttpRequestReplyClientSpec.java
```diff
@@ -123,6 +131,16 @@ private static Duration requireNonZeroDuration(Duration duration) {
     }
   }
 
+  private static final class DurationJsonSerialize extends JsonSerializer<Duration> {
```
sjwiesman, line 134: We may need to backport this commit. I'm not sure how RemoteFunctionBuilder ever worked...

statefun-flink/statefun-flink-datastream/src/main/java/org/apache/flink/statefun/flink/datastream/Endpoint.java
```diff
@@ -0,0 +1,163 @@
...
+ * available. Path templates may contain template parameters that are filled in based on the
+ * functions specific type. Template parameterization works well with load balancers and service
+ * gateways.
+ *
+ * <p>For example, a message sent to the typename {@literal 'com.example/greeter'} will be routed to
+ * {@literal 'http://bar.foo.com/greeter'}.
+ *
+ * <p>{@code Endpoint.withSpec("example/*", "https://endpoints/{function.name}"}</pre>
+ */
+public class Endpoint {
```
sjwiesman, line 55: I deprecated RemoteFunctionBuilder and replaced it with Endpoint mostly to align naming

statefun-flink/statefun-flink-datastream/pom.xml
```diff
@@ -76,12 +76,6 @@ under the License.
             <version>${flink.version}</version>
             <scope>provided</scope>
         </dependency>
-        <dependency>
-            <groupId>org.apache.flink</groupId>
-            <artifactId>flink-java</artifactId>
-            <version>${flink.version}</version>
-            <scope>provided</scope>
-        </dependency>
```
sjwiesman: As far as I can tell, this dependency was added by accident. Let me know if I should revert this commit.

sjwiesman: also curious why the dependency on `statefun-flink-io`?

sjwiesman:
rebased on #246

igalshilman:
@sjwiesman I'm closing this for now, because we will try to solve it without copying the sources, by properly shedding protobuf.

PR #246, opened 2021-07-29
[FLINK-23547][build] run integration tests on install

sjwiesman:
Stateful Functions is missing the surefire maven plugin and integration tests, marked with ITCase suffix, are not executed on install.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -318,12 +318,7 @@ under the License.
                     <trimStackTrace>false</trimStackTrace>
                     <systemPropertyVariables>
                         <forkNumber>0${surefire.forkNumber}</forkNumber>
-                        <checkpointing.randomization>true</checkpointing.randomization>
-                        <!-- on, unset, or random -->
-                        <checkpointing.changelog>unset</checkpointing.changelog>
                         <project.basedir>${project.basedir}</project.basedir>
-                        <!--suppress MavenModelInspection -->
-                        <test.randomization.seed>${test.randomization.seed}</test.randomization.seed>
                     </systemPropertyVariables>
                     <argLine>-Xms256m -Xmx2048m -Dmvn.forkNumber=${surefire.forkNumber} -XX:+UseG1GC</argLine>
                 </configuration>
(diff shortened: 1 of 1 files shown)


sjwiesman:
Well, I guess the question is was this on purpose? I did find at least one existing ITCase for Unix sockets that never runs on CI

pom.xml
```diff
@@ -307,6 +308,58 @@ under the License.
...
+            <!--surefire for unit tests and integration tests-->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>2.22.1</version>
+                <configuration>
+                    <trimStackTrace>false</trimStackTrace>
+                    <systemPropertyVariables>
+                        <forkNumber>0${surefire.forkNumber}</forkNumber>
+                        <checkpointing.randomization>true</checkpointing.randomization>
```
StephanEwen, line 321: This is Flink-specific, not needed in statefun.

pom.xml
```diff
@@ -307,6 +308,58 @@ under the License.
...
+                        <!-- on, unset, or random -->
+                        <checkpointing.changelog>unset</checkpointing.changelog>
```
StephanEwen, line 323: Same as above

pom.xml
```diff
@@ -307,6 +308,58 @@ under the License.
...
+                        <project.basedir>${project.basedir}</project.basedir>
+                        <!--suppress MavenModelInspection -->
+                        <test.randomization.seed>${test.randomization.seed}</test.randomization.seed>
```
StephanEwen, line 326: Same as above.

pom.xml
```diff
@@ -307,6 +308,58 @@ under the License.
...
+                        <phase>test</phase>
+                        <goals>
+                            <goal>test</goal>
+                        </goals>
+                        <configuration>
+                            <includes>
+                                <include>${test.unit.pattern}</include>
+                            </includes>
+                        </configuration>
+                    </execution>
```
StephanEwen, line 343: Maybe add `<reuseForks>true</reuseForks>` here to make it explicit.

StephanEwen (commented):
Looks good with some minor comments.

sjwiesman:
Thanks for the review, will address when merging

PR #249, opened 2021-08-10
[FLINK-23709] Remove SanityVerificationE2E and ExactlyOnceRemoteE2E

tzulitai:
It seems to be a general consensus that the smoke E2Es are extensive enough to subsume the `SanityVerificationE2E` and `ExactlyOnceRemoteE2E`, both of which are testing too specific scenarios and setups (e.g. a pre-defined order of very few events) to be useful in discovering critical bugs. As a matter of fact, a majority (if not all) of the more critical bugs were discovered solely by the smoke E2Es.

Since we're slowly growing in our build times, and the fact that it seems like the `SanityVerificationE2E` and `ExactlyOnceRemoteE2E` are not bringing much for our test coverage, we'd like to consider removing them completely.

Keep in mind, that in terms of test coverage, one could argue that by doing this we are losing out on test coverage against a running Kafka cluster for the Kafka ingress / egress.


igalshilman (approved):
+1 For removing the old tests, but I think that you have a good point about not having an e2e test that also covers Kafka.
What do you think about simplifying `sanity-e2e` to the bare minimum, and removing the `exactly-once` test which is subsumed by all the smoke tests?

tzulitai:
After an offline discussion, we've decided to merge this as is and add a Kafka E2E later.
Bottom line is, during release testing we should be giving the release a spin with the `flink-statefun-playground`, which includes applications that uses the Kafka ingress / egress.

PR #250, opened 2021-08-10
[FLINK-18810][sdk] Golang remote functions SDK

sjwiesman:
Implements a remote Golang SDK for stateful functions. 

# Important To Reviewers

`cell.go` is a bit tricky because it requires both resumable reads, as well as the ability to restart from the beginning. This is the component that was causing issues in testing so please take an extra close look. 

# Licensing

The SDK has two dependencies which are both BSD-3 licensed. Golang libraries do not have binary releases, only source releases, so my understanding is that we do not require a NOTICE file but please correct me if I am wrong. 

# Releasing 

The Go SDK is implicitly released when the release branch is tagged[1]. It does require the tag to be exactly the format `vMAJOR.MINOR.PATCH`, which is different than our current tagging model. We can discuss whether to add both tags or switch our tagging model during the release. 

[1] https://blog.golang.org/publishing-go-modules


tzulitai:
@sjwiesman We wont release any Go binaries, but the Go SDK source will be part of the source tarball release. Are the BSD-3 licensed dependencies bundled in the SDK source? If so, we need to acknowledge that here: https://github.com/apache/flink-statefun/blob/master/NOTICE. This is the NOTICE file that gets bundled with the source release.

tzulitai (commented):
Thanks for opening this PR @sjwiesman! I'm excited to see the StateFun Go SDK coming to shape.

I've only did a pass on the smoke E2E so far, please see my comments below:

statefun-sdk-go/v3/pkg/statefun/context.go
```diff
@@ -0,0 +1,170 @@
...
+
+	s.Lock()
+	s.response.DelayedInvocations = append(s.response.DelayedInvocations, invocation)
+	s.Unlock()
+}
+
+func (s *statefunContext) SendAfterWithCancellationToken(delay time.Duration, token string, message MessageBuilder) {
+	if len(token) == 0 {
+		panic("cancellation token cannot be empty")
+	}
```
austince, line 125: `token` is settable by the user, right? Could they handle this error case?

austince (commented):
Just taking a read through, looks like really nice work :)

statefun-sdk-go/v3/pkg/statefun/context.go
sjwiesman: Not really, but I think I can use a type alias to more eagerly validate the token.

igalshilman (approved):
Thanks for your contribution @sjwiesman!,  I'm sure that having a new SDK would benefit many users, and unlock new use cases!

*The SDK surface looks really ergonomic and feels smilier to our existing SDKS, so big +1 there!
* I'm happy to see that the smoke/verification tests are integrated and passing 👍 
* I've tried the SDK locally by:
1) stress testing a simple greeter via the harness
2) creating a StateFun app that has a Go function and a Python function, and having them message each other.
3) Checked that it is possible to send an egress message from a Go function.
4) Ran the smoke test locally with a high failure count. 

* I've also skimmed trough the main logic and few of the types, but as I'm not a Go dev, I can't really add much there. Can just say that it looks very familiar in structure to the Java and Python SDKS.

I think we should merge this (feel free to do it :-P) , and I will be continue testing in the following week with larger applications/different deployments/custom types.

One thing that we'd need to do as a followup is creating a walkthrough/greeter playground examples. 

Great work! 🥳

sjwiesman:
@igalshilman thanks for your help reviewing! Thrilled to finally get this merged in :)

PR #255, opened 2021-08-11
[FLINK-23718] [build, sdk] Manage Python SDK as a Maven module

tzulitai:
Previously, the StateFun Python SDK lives as a "dangling" directory in the repository that is not managed by Maven.

We'd like include the directory as a Maven module. To start things simple, we can have the Maven POM do nothing at all - the sole purpose is just so that the directory is included in the build process so we can take advantage of existing checks in the build process, e.g. for ASF license checks on the source files.

We can start extending on this in the future, e.g. build the SDK or run the Python unit tests using the Docker Maven plugin.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,6 +53,7 @@ under the License.
         <module>statefun-sdk-embedded</module>
         <module>statefun-sdk-protos</module>
         <module>statefun-sdk-java</module>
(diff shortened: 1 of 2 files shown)


igalshilman (approved):
Thanks! Looks good to me

PR #256, opened 2021-08-12
[FLINK-20574] [core] Block first request to mitigate excessive IncompleteInvocationContext roundtrips in RequestReplyFunction

tzulitai:
For the first request, we block until response is received; for stateful applications, especially at restore time of a restored execution where there may be a large backlog of events and checkpointed inflight requests, this helps mitigate excessive hoards of `IncompleteInvocationContext` responses and retry attempt round-trips.


tzulitai:
Some unit tests in`RequestReplyFunctionTest` is failing, am working at addressing those before this is merge-able.

It's mostly just poorly implemented UTs there and requires some attention to improve them.
Otherwise, the smoke E2Es are passing locally for me, so this is still a safe change that is worthwhile to consider.

tzulitai:
As discussed offline with @igalshilman, we're not going to add any extra unit test coverage for this change for now.
We workaround the unit tests by toggling the `isFirstRequestSent` flag to always be `true` in the current unit tests.

However, we should definitely revisit the `RequestReplyFunctionTest` sometime.
I've opened a JIRA to track that: https://issues.apache.org/jira/browse/FLINK-23762

tzulitai:
Merging this now ..

PR #258, opened 2021-08-18
[FLINK-23853] Update Flink dependency to v1.13.2

tzulitai:
Updates Flink dependency to v1.13.2, as well as update all relevant `NOTICE` files as a result of this upgrade.

The change is verified by the E2E tests.


igalshilman (approved):
Thanks @tzulitai, LGTM 👍

PR #265, opened 2021-09-01
[FLINK-24109] Correct egress string serialization

sjwiesman:
Fixes serialization of golang SDK egress builders to always use the utf-8 encoding of strings.

### DIFF
--- a/statefun-sdk-go/v3/pkg/statefun/egress.go
+++ b/statefun-sdk-go/v3/pkg/statefun/egress.go
@@ -20,8 +20,10 @@ import (
 	"bytes"
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"github.com/apache/flink-statefun/statefun-sdk-go/v3/pkg/statefun/internal/protocol"
 	"google.golang.org/protobuf/proto"
+	"unicode/utf8"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


statefun-sdk-go/v3/pkg/statefun/egress.go
```diff
@@ -78,7 +78,7 @@ func (k KafkaEgressBuilder) toEgressMessage() (*protocol.FromFunction_EgressMess
 	} else {
 		switch value := k.Value.(type) {
 		case string:
-			_ = StringType.Serialize(&buffer, value)
+			buffer.WriteString(value)
```
igalshilman, line 81: Does it make sense to size the `buffer` before writing the string?
I can imagine folks sending to en egress large json blobs.

igalshilman, line 81: I think that there is an encoding issues here, as I understand strings in Go are not necessarily `utf-8` encoded.
(constants are utf8, while at runtime they are not necessarily)
Do you think that this is an issue worth addressing?

Perhaps at least we need to add a comment in the method documentation? to use to correct string encoding to match the expected encoding in the Kafka topic. 
This is a slight difference from the Python and Java SDKs as there we also auto deduce strings to be utf-8 strings.

igalshilman (approved):
Thanks for addressing this issue @sjwiesman.
I have one comment regarding string encoding, other then that looks good to me 👍 

I also assume that you've inspected the output manually in the greeter example, just as a sanity check (that what surfaced the issue for me)

statefun-sdk-go/v3/pkg/statefun/egress.go
sjwiesman: You are correct, I added an explicit utf-8 encoding step.

sjwiesman: The encoding change made this unnecessary, see other comment.

sjwiesman:
rebased, will merge on green.

PR #273, opened 2021-10-06
[FLINK-24464][core] Metrics sometimes report negative backlog

sjwiesman:
The backlog metric sometimes reports a negative value. This is because the current count is stored in an in-memory variable, while the value used to decrement the count as the backlog clears is stored in Flink state. In the case of a job restart, the in-memory variable is reset to zero (it is a new instance) while the numbers in state are retained. 

Reseting the count after a restore would require iterating over all keys, so instead we simply prevent negative backlogs from being reported to not publish non-sensical values to reporters.

### DIFF
--- a/statefun-flink/statefun-flink-core/src/test/java/org/apache/flink/statefun/flink/core/metrics/NonNegativeCounterTest.java
+++ b/statefun-flink/statefun-flink-core/src/test/java/org/apache/flink/statefun/flink/core/metrics/NonNegativeCounterTest.java
@@ -1,0 +1,27 @@
+package org.apache.flink.statefun.flink.core.metrics;
+
+import static org.hamcrest.CoreMatchers.is;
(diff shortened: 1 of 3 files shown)


statefun-flink/statefun-flink-core/src/test/java/org/apache/flink/statefun/flink/core/metrics/NonNegativeCounterTest.java
```diff
@@ -0,0 +1,27 @@
+package org.apache.flink.statefun.flink.core.metrics;
```
igalshilman, line 1: Missing header

igalshilman (approved):
Thanks Seth, LGTM

sjwiesman:
I'll fix the header while merging, grazi!

PR #282, opened 2021-12-26
[FLINK-25197] Fix serialization issue in RequestReplyFunctionBuilder

galenwarren:
This fixes the issue described in [FLINK-25197](https://issues.apache.org/jira/browse/FLINK-25197), where this error occurs when using RequestReplyFunctionBuilder to build Statefun jobs:

> Java 8 date/time type `java.time.Duration` not supported by default: add Module "org.apache.flink.shaded.jackson2.com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling 

The fix is to make RequestReplyFunctionBuilder use an instance of StatefunObjectMapper to serialize the client spec -- previously, it used a non-customized instance of ObjectMapper -- and to change StatefunObjectMapper to support serialization of java.time.Duration values.


tillrohrmann (approved):
Thanks a lot for this PR @coachclientconnect. Your fix is spot on. I will address my comments while merging this PR.

galenwarren:
@tillrohrmann Thanks!

PR #283, opened 2022-01-19
[FLINK-25708] Bump Flink dependency to 1.14.3

tillrohrmann:
This PR bumps the Flink dependency to 1.14.3. I had to add some dependency management entries to resolve conflicting transitive dependencies.

Optional: This PR also contains a very simple migration test based on the `EmbeddedSmokeHarnessTest`. This test is not ignored as the other `Harness` tests are.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,8 +79,11 @@ under the License.
         <protobuf.version>3.7.1</protobuf.version>
         <unixsocket.version>2.3.2</unixsocket.version>
         <protoc-jar-maven-plugin.version>3.11.1</protoc-jar-maven-plugin.version>
-        <flink.version>1.13.2</flink.version>
+        <flink.version>1.14.3</flink.version>
         <scala.binary.version>2.12</scala.binary.version>
+        <scala.version>2.12.7</scala.version>
+        <lz4-java.version>1.8.0</lz4-java.version>
+        <flink-shaded-jackson.version>2.12.4-14.0</flink-shaded-jackson.version>
         <test.unit.pattern>**/*Test.*</test.unit.pattern>
     </properties>
 
--- a/statefun-e2e-tests/pom.xml
+++ b/statefun-e2e-tests/pom.xml
@@ -44,6 +44,28 @@ under the License.
         <module>statefun-smoke-e2e-js</module>
     </modules>
 
+    <dependencyManagement>
+        <dependencies>
+            <!--
+            Pin the scala library version in order to resolve the dependency conversion problem between two
+            transitive scala-library versions in flink-scala
+            -->
+            <dependency>
+                <groupId>org.scala-lang</groupId>
+                <artifactId>scala-library</artifactId>
+                <version>${scala.version}</version>
+            </dependency>
+            <!--
+            Pin version to avoid conflicts between flink-runtime and kafka-clients
+            -->
+            <dependency>
+                <groupId>org.lz4</groupId>
+                <artifactId>lz4-java</artifactId>
+                <version>${lz4-java.version}</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
     <build>
         <plugins>
             <plugin>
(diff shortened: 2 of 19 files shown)


statefun-flink/statefun-flink-harness/src/main/java/org/apache/flink/statefun/flink/harness/Harness.java
```diff
@@ -126,6 +129,14 @@ public Harness withSavepointLocation(String savepointLocation) {
...
 
   public void start() throws Exception {
+    try {
+      startAsync().getJobExecutionResult().get();
+    } catch (ExecutionException ee) {
+      ExceptionUtils.rethrowException(ExceptionUtils.stripExecutionException(ee));
+    }
+  }
+
+  public JobClient startAsync() throws Exception {
```
igalshilman, line 139: I'm not sure that I'd like to expose the `JobClient` as a public method on the `Harness`.
How about we think about it as a followup PR?

statefun-e2e-tests/statefun-smoke-e2e-embedded/src/test/java/org/apache/flink/statefun/e2e/smoke/embedded/EmbeddedSmokeHarnessMigrationTest.java
```diff
@@ -0,0 +1,127 @@
...
+import org.apache.flink.statefun.e2e.smoke.SmokeRunnerParameters;
+import org.apache.flink.statefun.flink.harness.Harness;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@RunWith(Parameterized.class)
+public class EmbeddedSmokeHarnessMigrationTest {
```
igalshilman, line 39: Thanks for adding a migration test!
This test would catch any internal migration issues with the core statefun job graph.

Perhaps we need to followup with a separate e2e migration test, that runs within its own migration profile, that includes at least a Kafka ingress/egress pair.
It can be completely based remote functions, without any verification/failure injection aspects to it, and this probably should be based on `testcontainers` instead of the harness.

having test containers will allows the test to create a save point in an arbitrary previous version (say 3.1.1) 
based on the previously released official docker images, and load it with the current version.
The test would be bit more time consuming, and that's why I would recommend that to run under a different profile.

If this is something you find reasonable, then I'll create a followup JIRA issue.

igalshilman (approved):
Thanks @tillrohrmann for migrating StateFun for the latest Flink version!
I've run the end to end tests locally, and they pass.

I have one comment about exposing Flink`s `JobClinet` in the `Harness` , for your consideration, and a followup regarding a migration test.

Thanks again!

statefun-flink/statefun-flink-harness/src/main/java/org/apache/flink/statefun/flink/harness/Harness.java
tillrohrmann: Makes sense.

statefun-e2e-tests/statefun-smoke-e2e-embedded/src/test/java/org/apache/flink/statefun/e2e/smoke/embedded/EmbeddedSmokeHarnessMigrationTest.java
tillrohrmann: I think this sounds like a good idea. Having these test to ensure that nothing is breaking unintentionally is a good improvement. I will remove the commit that introduced the harness test and we continue based on the migration test JIRA ticket you'll create.

tillrohrmann:
Thanks a lot for your review @igalshilman. I've removed the migration test commit and rebased onto the latest master. Will merge it once GHA gives green light.

I'd be +1 for following up with a dedicated JIRA ticket for adding a migration test. It will help spotting incompatible changes faster.

cosmir17:
The pr has been merged 🎉🥳 I should just wait. Any news on public dependency release? @tillrohrmann 😅

tillrohrmann:
I will propose to create a new release today. So expect that the release happens later this week.

cosmir17:
Thank you for your kindness 🥳🥰 @tillrohrmann

tillrohrmann:
@cosmir17 if you want to help with the release verification, then I can ping you once we have the release artifacts created.

PR #285, opened 2022-01-26
[FLINK-25828] Let build-stateful-functions.sh ignore javadoc and sources jar

tillrohrmann:
This commit lets the build-stateful-functions.sh script ignore javadoc and sources jars that
can make the following copy commands to fail.


igalshilman (approved):
Thanks Till, this looks like a good addition!

tillrohrmann:
Thanks for the review Igal. Merging this PR now.

PR #289, opened 2022-01-27
[hotfix] remove invalid link in README

mingmxu:
`statefun-examples` is not there anymore, remove it from README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,6 @@ to get yourself started with Stateful Functions.
 
 For a fully detailed documentation, please visit the [official docs](https://ci.apache.org/projects/flink/flink-statefun-docs-master).
 
-For code examples, please take a look at the [examples](statefun-examples/).
-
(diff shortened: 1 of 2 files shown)


igalshilman:
Thank you for your contribution @mingmxu,  and welcome aboard :-)

Thats a good catch, indeed we don't have the examples there anymore. But we do have the following repo:

https://github.com/apache/flink-statefun-playground

What do you think about changing the link to point to the playground repo?

mingmxu:
> What do you think about changing the link to point to the playground repo?
That's a good idea. The question is that version of `flink-statefun-playground` is behind. For long term, should we move it back to repo `flink-statefun` to align the snapshot/release version?

mingmxu:
I forgot that `flink-statefun-playground` is included in chapter `Getting Started`, prefer to just remove the line here.

Version of `flink-statefun-playground` is another story ATM.

igalshilman:
Thanks a lot @mingmxu! merged.

PR #300, opened 2022-02-09
[FLINK-25933][statefun] Add support for the async HTTP transport in request-reply function builder

galenwarren:
The objective here is to allow users who build Statefun jobs in code -- using RequestReplyFunctionBuilder and related classes -- to choose the newer, async HTTP transport for remote function class. Previously, only the sync HTTP transport was supported.

The synchronous and asynchronous transports have different sets of configurable properties -- neither of which is a subset of the other -- so the existing `RequestReplyFunctionBuilder` continues to be used to build sync functions, and the new `AsyncRequestReplyFunctionBuilder` is used to build async functions. Each of these derive from a new common base class, `StatefulFunctionBuilder`.

`StatefulFunctionBuilder` has factory methods `requestReplyFunctionBuilder` and `asyncRequestReplyFunctionBuilder` to create sync and async builders, and the factory method `requestReplyFunctionBuilder` on `RequestReplyFunctionBuilder` is deprecated.

The `SerializableHttpFunctionProvider` class is updated to properly choose the transport client based on the spec; previously, it always chose the sync client.

Added unit tests and updated docs.

### DIFF
--- a/docs/content/docs/sdk/flink-datastream.md
+++ b/docs/content/docs/sdk/flink-datastream.md
@@ -74,6 +74,7 @@ as DataStream egresses to produce the outputs to:
 ```java
 FunctionType GREET = new FunctionType("example", "greet");
 FunctionType REMOTE_GREET = new FunctionType("example", "remote-greet");
+FunctionType REMOTE_GREET_ASYNC = new FunctionType("example", "remote-greet-async");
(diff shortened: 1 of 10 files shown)


igalshilman (approved):
Thanks @galenwarren! the changes look good to me 👍

PR #301, opened 2022-02-10
[FLINK-25898] Add README.md to statefun-sdk-js

tillrohrmann:
This commit adds an README.md to statefun-sdk-js. This readme will be packaged with the npm package and published to the npm registry (https://www.npmjs.com/package/apache-flink-statefun).

### DIFF
--- a/statefun-sdk-python/README.md
+++ b/statefun-sdk-python/README.md
@@ -2,11 +2,11 @@
 
 Stateful Functions is an API that simplifies the building of **distributed stateful applications** with a **runtime built for serverless architectures**.
 It brings together the benefits of stateful stream processing - the processing of large datasets with low latency and bounded resource constraints -
-along with a runtime for modeling stateful entities that supports location transparency, concurrency, scaling, and resiliency. 
+along with a runtime for modeling stateful entities that supports location transparency, concurrency, scaling, and resiliency.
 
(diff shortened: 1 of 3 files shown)


statefun-sdk-python/README.md
```diff
@@ -17,7 +17,7 @@ to get yourself started with Stateful Functions in Python.
 
 For a fully detailed documentation, please visit the [official docs](https://ci.apache.org/projects/flink/flink-statefun-docs-master).
 
-For code examples, please take a look at the [examples](../statefun-examples/).
+For code examples, please take a look at the [examples](https://github.com/apache/flink-statefun-playground/tree/dev/python).
```
igalshilman, line 20: Can that not be pointing to `dev` but rather to the main branch, or the latestest released branch?

igalshilman (approved):
Thanks for adding the readme, @tillrohrmann !
The readme reads well imho, I have one little comment about the `dev` url for your consideration.

statefun-sdk-python/README.md
tillrohrmann: I think the latest release branch makes sense since it is also what is shown for a specific PyPI package.

tillrohrmann:
Thanks for the review @igalshilman. I will address your comment and then merge this PR.

statefun-sdk-python/README.md
tillrohrmann: What is not so optimal is that updating these links is currently a manual process. As a follow-up we might wanna think about how this could be automated by establishing a branch mapping between `flink-statefun` and `flink-statefun-playground`.

PR #303, opened 2022-02-23
[FLINK-26340][statefun-golang-sdk] Add ability in Golang SDK to create new statefun.Context from existing one, but with a new underlying context.Context

galenwarren:
In the Golang SDK, `statefun.Context` embeds the `context.Context` interface and is implemented by the `statefunContext` struct, which embeds a `context.Context`. To support common patterns in Golang related to adding values to context, it would be useful to be able to create a derived `statefun.Context` that is equivalent to the original in terms of statefun functionality but which wraps a different `context.Context`.

This PR adds:

```
WithContext(ctx context.Context) statefun.Context
```

... to the `statefun.Context` interface and implements it on `statefunContext`. This method returns the derived statefun context.

This is technically a breaking change to `statefun.Context` (i.e. addition of new method), but, given its purpose, we do not expect there to be implementations of this interface outside the Golang SDK. 

The changes here include:
* Change `sync.Mutex` in `statefunContext` to be `*sync.Mutex` and update construction of statefunContext accordingly
* Add `WithContext` to `statefun.Context` interface and implement on `statefunContext`
* Add unit test

### DIFF
--- a/statefun-sdk-go/v3/pkg/statefun/context.go
+++ b/statefun-sdk-go/v3/pkg/statefun/context.go
@@ -18,9 +18,11 @@ package statefun
 
 import (
 	"context"
-	"github.com/apache/flink-statefun/statefun-sdk-go/v3/pkg/statefun/internal/protocol"
+	"errors"
 	"sync"
 	"time"
+
+	"github.com/apache/flink-statefun/statefun-sdk-go/v3/pkg/statefun/internal/protocol"
 )
 
 // A Context contains information about the current function invocation, such as the invoked
@@ -64,7 +66,7 @@ type Context interface {
 }
 
 type statefunContext struct {
-	sync.Mutex
+	*sync.Mutex
 	context.Context
 	self     Address
 	caller   *Address
@@ -159,4 +161,17 @@ func (s *statefunContext) SendEgress(egress EgressBuilder) {
 	s.Lock()
 	s.response.OutgoingEgresses = append(s.response.OutgoingEgresses, msg)
 	s.Unlock()
+}
+
+// DeriveContext derives a new statefun.Context from an existing one, replacing
+// the wrapped context.Context.
+func DeriveContext(statefunCtx Context, ctx context.Context) Context {
+	switch value := (statefunCtx).(type) {
+	case *statefunContext:
+		newStatefunContext := *value
+		newStatefunContext.Context = ctx
+		return &newStatefunContext
+	default:
+		panic(errors.New("stateful function context supplied to DeriveContext is not recognized"))
+	}
 }
--- a/docs/content/docs/sdk/golang.md
+++ b/docs/content/docs/sdk/golang.md
@@ -323,6 +323,27 @@ func (g *Greeter) Invoke(ctx statefun.Context, message: statefun.Message) error 
 {{< /tab >}}
 {{< /tabs >}}
 
+## Context
+
+The `Context` interface exposed by the Golang SDK -- which is used above to access storage, egresses, and invoke other stateful functions -- embeds the standard Golang `Context` interface from the `context` package. You can further customize the wrapped `Context` using `DeriveContext`, for example:
+
+```
+import (
+    "github.com/apache/flink-statefun/statefun-sdk-go/v3/pkg/statefun"
+)
+
+func (g *Greeter) Invoke(ctx statefun.Context, message: statefun.Message) error {
+
+    ctx = statefun.DeriveContext(ctx, context.WithValue(ctx, "key", "value"))
+
+    // do something with ctx, which now holds key=value
+
+    return nil
+}
+```
+
+`DeriveContext` accepts a stateful-function `Context` and a standard `Context`; it returns a new stateful-function `Context` that is functionally equivalent to the original stateful-function `Context`, as far as stateful-function operations are concerned, but which wraps the supplied standard `Context` instead. 
+
 ## Serving Functions
 
 The Golang SDK ships with a ``RequestReplyHandler`` that is a standard http `Handler` and automatically dispatches function calls based on RESTful HTTP ``POSTS``.
--- a/statefun-sdk-go/v3/pkg/statefun/context_test.go
+++ b/statefun-sdk-go/v3/pkg/statefun/context_test.go
@@ -17,11 +17,23 @@
 package statefun
 
 import (
+	"context"
+	"sync"
+	"testing"
+	"time"
+
 	"github.com/apache/flink-statefun/statefun-sdk-go/v3/pkg/statefun/internal/protocol"
 	"github.com/stretchr/testify/assert"
 	"google.golang.org/protobuf/proto"
-	"testing"
-	"time"
+)
+
+type testContextKeyType string
+
+const (
+	testContextKey1   = testContextKeyType("key1")
+	testContextValue1 = "value1"
+	testContextKey2   = testContextKeyType("key2")
+	testContextValue2 = "value2"
 )
 
 func TestStatefunContext_Send(t *testing.T) {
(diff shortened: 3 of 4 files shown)


statefun-sdk-go/v3/pkg/statefun/context.go
```diff
@@ -76,6 +81,12 @@ func (s *statefunContext) Storage() AddressScopedStorage {
 	return s.storage
 }
 
+func (s *statefunContext) WithContext(ctx context.Context) Context {
+	newContext := *s
+	newContext.Context = ctx
+	return &newContext
```
austince, line 87: Will this then share the mutex across both contexts? Reading [this thread](https://groups.google.com/g/golang-nuts/c/imxjBLNJ9OY), it seems like another approach would be to lock the mutex, copy the protected fields into a new object (which holds a new mutex), then unlock. We may also want to copy the other pointer fields like caller, storage, address. wdyt?

galenwarren: Unless I'm mistaken, the initial assignment operation:

```
newContext := *s
```

... will shallow copy all the fields from the existing `statefunContext` to the new one, including the caller (a pointer), self (not a pointer), storage (a pointer), address (a pointer), response (a pointer), and the mutex (a pointer). 

This means that both contexts will share the same mutex and response, which is what we want, I think? The mutex is used to serialize access to fields in the response, i.e.:

```
s.Lock()
s.response.OutgoingMessages = append(s.response.OutgoingMessages, invocation)
s.Unlock()
```

My read of the linked thread is that it describes how to copy a mutex when it is directly embedded (i.e. not a pointer); in this case, we have the option to make it a pointer and sidestep the problem. But please let me know if you disagree. Thanks.

austince: Ah yeah, I was thinking we want this to be a full, independent copy of the Context, without shared state (like in `http.Request`). I don't have enough "context" here to know if that's a requirement / what possible situations would arise if the state is shared.

austince: @sjwiesman ?

galenwarren: Ah, I understand what you're saying.

I think that, with regard to the mutex and the response, we need the old and new contexts to share the same state, because both contexts need to update the same response object, i.e. the one (and only) response object that will be returned from the invocation of the stateful function.

I think that also holds true for the storage; there is only one storage instance that applies for the entire stateful-function call, so both contexts need to reference it. (And the storage operations are protected with a sync.RWMutex, so this should be safe.)

So that just leaves the caller. My understanding is that this is a pointer in order that it can be nil, to handle the case where there is no caller, but that it is also intended to be read-only. Even without the addition of `WithContext`, two consumers of the same statefun.Context could conceivably step on each others' toes by modifying the `Address` returned from `Caller`. I think the assumption is that the returned `Address` will be read-only, and with that assumption, we wouldn't be making anything worse by shallow copying vs. deep copying.

But, deep copying the caller `Address` would certainly be possible.

@sjwiesman @tillrohrmann Do you have any thoughts on this?

galenwarren:
@austince @tillrohrmann @sjwiesman 

It just occurred to me that there **is** a circumstance where people might reasonably be implementing the statefun.Context interface -- for unit testing. So this change could break unit tests where statefun.Context is mocked :(.

Would that be sufficient reason to leave statefun.Context alone and make this a top-level function? All things being equal, I think it's cleaner as-is, but all things may not be equal.

statefun-sdk-go/v3/pkg/statefun/context.go
sjwiesman: The statefun context very explicitly has shared state. It is effectively building the return payload from the batch of messages. Self is static for all messages in the batch. Caller is message specific, but from user code perspective both are immutable for the full invocation.

austince (approved):
`DeriveContext` seems like a nice name to me + rest looks good too :tada:

tillrohrmann (approved):
From what I can tell the changes look good to me. Thanks a lot for providing this PR @galenwarren. Maybe a final approval from @sjwiesman would be nice before we merge this change.

sjwiesman (changes requested):
I'm concerned that this change prevents us from ever being able to add our own kv pairs to the context. What about something like this where we union them together. 

```go
type statefunContext struct {
	sync.Mutex
	context.Context
	merged   []context.Context
	self     Address
	caller   *Address
	storage  *storage
	response *protocol.FromFunction_InvocationResponse
}

func (s *statefunContext) AddAll(other context.Context) {
	s.Lock()
	s.merged = append(s.merged, other)
	s.Unlock()
}

func (s *statefunContext) Value(key interface{}) interface{} {
	for i := len(s.merged)-1; i >= 0; i-- {
		if result := s.merged[i].Value(key); result != nil {
			return result
		}
	}
	return s.Context.Value(key)
}
```

statefun-sdk-go/v3/pkg/statefun/context.go
```diff
@@ -64,7 +66,7 @@ type Context interface {
 }
 
 type statefunContext struct {
-	sync.Mutex
+	*sync.Mutex
```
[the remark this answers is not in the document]
galenwarren: This was changed to a pointer so that different statefunContext instances can reference the same sync.Mutex, as opposed to each having their own sync.Mutex. I thought this was necessary because derived statefunContext instances all reference the same response object, and the mutex is used to prevent concurrent modifications to the response object?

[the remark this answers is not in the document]
sjwiesman: You are correct. Thank you.

sjwiesman:
By "us" I am referring to the StateFun SDK. Let's suppose in a future version we add the pair `a=b` as a default bit of information to the context[^1]. In the current implementation, where we are deriving a new context with the pair `c=d`, that new `statefun.Context` will lose the original pair `a=b`. This behavior differs from `context.Context.WithValue` which retains the existing kv pairs when new ones are added. Maybe my example code should be changed to create a new statefun context that contains the underlying contexts and the new one, something like: 

```go
type statefunContext struct {
	*sync.Mutex
	context.Context
	merged   []context.Context
	self     Address
	caller   *Address
	storage  *storage
	response *protocol.FromFunction_InvocationResponse
}

func (s *statefunContext) WithAllValues(other context.Context) Context {
	s.Lock()
	defer s.Unlock()
	return &statefunContext{
		s.Mutex,
		s.Context,
		append(s.merged, other),
		s.self,
		s.caller,
		s.storage,
		s.response,
	}
}

func (s *statefunContext) Value(key interface{}) interface{} {
	for i := len(s.merged) - 1; i >= 0; i-- {
		if result := s.merged[i].Value(key); result != nil {
			return result
		}
	}
	return s.Context.Value(key)
}
```

While I realize in go you can at any point create a new context that discards all underlying data, that doesn't seem like a usage pattern we want to encourage.

[^1]: A real example, I can imagine adding tracing information into the invocations in the near future.

galenwarren:
@sjwiesman 

> Let's suppose in a future version we add the pair a=b as a default bit of information to the context[1](https://github.com/apache/flink-statefun/pull/303#user-content-fn-1-51415ccdcff656cb887c25acf61bd7df). In the current implementation, where we are deriving a new context with the pair c=d, that new statefun.Context will lose the original pair a=b

I don't think this would be the case in the current implementation. Consider:

```
ctx = statefun.DeriveContext(ctx, context.WithValue(ctx, "key", "value"))
```

... but expanded out a bit:

```
func Invoke(ctx statefun.Context, message statefun.Message) error {

	// assume ctx contanis a=b when Invoke is called
	newInnerCtx := context.WithValue(ctx, "c", "d") 	// newInnerCtx ctx holds a=b and c=d
	newCtx := DeriveContext(ctx, newInnerCtx)		// newCtx holds a=b and c=d
}
```

The key is that the `context.Context` that is passed as the second parameter to `DeriveContext` itself "extends" the original `statefun.Context`, `ctx`. So it contains all kv pairs from that original context, plus the new one.

I modified the unit test to demonstrate this, by preloading the initial `statefun.Context` with a kv pair and verifying that the kv pair is present in the derived context.

sjwiesman:
@austince I don't mean internal metadata, I mean information we want to make available to users. If we have tracing keys to see how messages are passed through the systems, functions may want access to those if they are going to call out to external apis themselves. 

Maybe I'm having trouble following the code but I don't understand how this works in the current implementation, because right [here](https://github.com/apache/flink-statefun/pull/303/files#diff-48c8d3a43a8c7e2bcb04e3033333fffe1cf878e94745f6b28130328bd6d151fdR172) we are overriding the original context.

```go
// in the library
tracingKey := "trace"
traceID := "ABC"
originalContext := StatfunContext(context.WithValue(ctx, tracingKey, traceId)) // fake, internally constructed initial context


// in the Invoke method, user code
newContext := DeriveContext(originalContext, context.WithValue(originalContext, "a", "b"))

newContext.Value(tracingKey) == "ABC"
newContext.Context != originalContext.Context
```
I'm also curious whose `Done` method will be respected. The one from the original HTTP request or the new context we derived from?

sjwiesman:
> The key is that the context.Context that is passed as the second parameter to DeriveContext itself "extends" the original statefun.Context, ctx. So it contains all kv pairs from that original context, plus the new one.

Can you please explain to me how this works? Why isn't derive just a variable reassignment as it looks in the code? I also have the question about `Done`.

sjwiesman:
I apologize, I missed the line in your example where we expect the new context to be built from the original statefun context in user code. This feels like more of a philosophical design question. Do we want to support replacing the root context (your implementation) of the statefun context or only extending it (mine). I'm not sure what the correct answer is here but I won't block the PR on it.

galenwarren:
> Can you please explain to me how this works?

I think it's just the [behavior](https://pkg.go.dev/context#WithValue) of `context.WithValue`, i.e. 

> WithValue returns a copy of parent in which the value associated with key is val.

So the new context is a copy of the original -- all the kv pairs, the Done channel, etc., but just with the new kv pair added. So these extended contexts form a tree structure, i.e. each child context uses its parent as the starting point.

> Why isn't derive just a variable reassignment as it looks in the code?

In an `Invoke` method, we just have access to the interface `statefun.Context`, but an assignment of Context would have to happen on the non-exported type `statefunContext` that implements that interface.  

The reason to replace the context completely -- as opposed to, say, allowing one to assign kv pairs in `statefun.Context`  -- is that, often, the context keys are non-exported types, and the only way to put a new value into context is via an exported function that returns a `context.Context`. An example would be the `WithLogger` function, [here](https://pkg.go.dev/knative.dev/pkg/logging#WithLogger).

galenwarren:
>  This feels like more of a philosophical design question. Do we want to support replacing the root context (your implementation) of the statefun context or only extending it (mine).

Agreed.

In my opinion, allowing the inner `statefun.Context` to be reassigned to *any* `context.Context` value  -- even one that has no relationship to the original, inner context -- is consistent with how one can choose to send *any* `context.Context` to a downstream Go function -- even one that has no relationship to whatever `context.Context` might have been received by the function -- in a general situation outside of statefun.

Of course, this would rarely be done, because the whole point of context is that it should flow from function call to function call, and generally all one does is add new values to an existing context and pass it along. But in general, one can choose to pass any context downstream, and I don't see why we would need to be more restrictive in statefun.

Just my two cents, though. What do you think?

sjwiesman (approved):
Spoke with @igalshilman offline and we are going to go with your proposal. Thank you for the PR!

galenwarren:
@sjwiesman Great! Thanks for taking the time to review this. I'll go ahead and squash the commits.

galenwarren:
Squashed. Is there anything else I need to do on this one? Thanks again.

PR #304, opened 2022-02-25
[FLINK-26375][statefun-golang-sdk] Fix Statefun Golang SDK to return nil from Context.Caller when there is no caller

galenwarren:
When a stateful function is invoked from an ingress – i.e. when there is no upstream function caller – Context.Caller() should return nil. Currently, it returns a pointer to a zero-value Address. This PR fixes that.

Changes:
* Change `invoke` method in `handler.go` to only populate statefunContext.caller when a caller is present in the invocation
* Add unit tests

### DIFF
--- a/statefun-sdk-go/v3/pkg/statefun/handler.go
+++ b/statefun-sdk-go/v3/pkg/statefun/handler.go
@@ -225,11 +225,10 @@ func (h *handler) invoke(ctx context.Context, toFunction *protocol.ToFunction) (
 			var cancel context.CancelFunc
 			sContext.Context, cancel = context.WithCancel(ctx)
 
-			var caller Address
 			if invocation.Caller != nil {
-				caller = addressFromInternal(invocation.Caller)
+				caller := addressFromInternal(invocation.Caller)
+				sContext.caller = &caller
 			}
-			sContext.caller = &caller
 			msg := Message{
 				target:     batch.Target,
 				typedValue: invocation.Argument,
--- a/statefun-sdk-go/v3/pkg/statefun/handler_test.go
+++ b/statefun-sdk-go/v3/pkg/statefun/handler_test.go
@@ -1,0 +1,89 @@
+package statefun
+
+import (
+	"context"
+	"testing"
+
+	"github.com/apache/flink-statefun/statefun-sdk-go/v3/pkg/statefun/internal/protocol"
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/protobuf/proto"
+)
+
+// helper to create a protocol Address from an Address
+func toProtocolAddress(address *Address) *protocol.Address {
+	if address != nil {
+		return &protocol.Address{
+			Namespace: address.FunctionType.GetNamespace(),
+			Type:      address.FunctionType.GetType(),
+			Id:        address.Id,
+		}
+	} else {
+		return nil
+	}
+}
+
+// helper to create a handler and invoke the function
+func invokeStatefulFunction(ctx context.Context, target *Address, caller *Address, argument *protocol.TypedValue, statefulFunction StatefulFunction) error {
+
+	builder := StatefulFunctionsBuilder()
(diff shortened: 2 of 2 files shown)


statefun-sdk-go/v3/pkg/statefun/handler.go
```diff
@@ -225,11 +225,10 @@ func (h *handler) invoke(ctx context.Context, toFunction *protocol.ToFunction) (
 			var cancel context.CancelFunc
 			sContext.Context, cancel = context.WithCancel(ctx)
 
-			var caller Address
 			if invocation.Caller != nil {
```
igalshilman, line 228: Can you please double check that this condition is enough? 
as far as I know in Protobuf there are never nil fields, a missing field would always be deserialized to an empty default value.
You can check the JavaSDk where we consider an empty Address to be missing.

igalshilman: Perhaps in the Protobuf Go it is somehow done differently, I don't know.

galenwarren: If I'm thinking about it right, I think the new unit tests demonstrate this is working, i.e. they:
* Create an instance of `protocol.ToFunction`, which contains (among other things) the invocations and associated callers (with the caller populated in one test and not populated in another test)
* Serialize that `protocol.ToFunction` to a `[]byte` using `proto.Marshal`
* Call `handler.Invoke` passing the serialized `protocol.ToFunction`, which internally deserializes the `protocol.ToFunction` using `proto.Unmarshal` before continuing to process it
* Validates that the resulting `statefunContext.caller` (via Context.Caller) is nil or a not-nil value, according to whether a caller was supplied in the original `protocol.ToFunction`

So, if a nil caller in the `protocol.ToFunction` were getting deserialized to a non-nil value via `proto.Unmarshal`, that would cause the `TestStatefunHandler_WithNoCaller_ContextCallerIsNil` unit test to fail.

Separately, I do a reasonable amount of work with Golang and protobuf, and while it's true that, in protobuf 3, primitive fields in messages are not deserialized to nil, embedded *message* fields can be deserialized to nil, as can repeated values where there are no values. The general rule for protobuf in Golang, as I understand it, is that fields not specified in a protobuf message are deserialized to the appropriate zero value in the generated Golang structs; since nested protobuf messages are generated as *pointers* to structs, the zero value used in those cases is nil.

There's some discussion of this [here](https://github.com/golang/protobuf/issues/15); in particular, note:

> When the .proto file specifies syntax="proto3", there are some differences:  Non-repeated fields of non-message type are values instead of pointers.

igalshilman: The test indeed demonstrates that it works, I was a bit surprised initially because with `Protobuf` 3 it should be impossible to differentiate between a missing field or a field that is set to default values (even an embedded one).
But your explanation, with a pointer being defaulted to `nil` makes sense! 👍
So let's go with that :-) !

igalshilman (approved):
Thanks @galenwarren!
The changes looks good to me, thanks for spotting and fixing this issue!

statefun-sdk-go/v3/pkg/statefun/handler.go
galenwarren: Thanks! I appreciate your taking the time to do the review.

Fwiw, I think you'll also see methods in Java-generated protobuf code that would allow you to determine if fields are set, but only for fields that correspond to messages. For example, here's one from my project:

```
    /**
     * <code>.google.protobuf.Timestamp updated_at = 3;</code>
     * @return Whether the updatedAt field is set.
     */
    @java.lang.Override
    public boolean hasUpdatedAt() {
      return updatedAt_ != null;
    }
```

... which is generated for a message declared as follows:

```
message SomeMessage {
  google.protobuf.Timestamp updated_at = 3;
}
```

PR #306, opened 2022-03-02
[FLINK-25866][statefun] Support additional TLS configuration

FilKarnicki:
### What is the purpose of the change

The goal of this PR is to add mutual TLS capabilities to flink-statefun. In Master, currently only server-side TLS is allowed, meaning the flink-statefun job can assert whether the remote service it's talking to is who it says it is, but not the other way around. Furthermore, only the default java certificate authority truststore can currently be used.

This change allows users to specify additional settings in module.yaml's function spec:
```yaml
spec:
  transport:
    type: io.statefun.transports.v1/async
    #(...)
    trust_cacerts: ~/trustedCAs.pem
    client_cert: classpath:clientPublic.crt
    client_key: ~/clientPrivate.key
    client_key_password: changeme
```
Certs/keys get loaded with `ResourceLocator` and used in Netty's `SslContext`

### Main changes are:

- Added a test for `NettyClient`
- Made slight changes to `HttpConnectionPoolManager` in order to catch failure case responses without them being automatically retried
- Included a bunch of certificates/keys for testing (with README.md files explaining how the certs were created)
- Replaced undertow in `statefun-smoke-e2e-java` with `netty` in order to be able to use the same cert/key loading mechanisms as the netty client in `statefun-flink-core`

### Verifying this change
`TransportClientTest` launches a netty service returning a stub response. There are three endpoints: http, https with required mutual TLS and https with server-side TLS only (for verifying existing, default jre truststore usecases)

`NettyClientTest` contains a number of tests used to verify the correct behaviour

`statefun-smoke-e2e-java` was updated to use mutual TLS

The public API, i.e., is any changed class annotated with @Public(Evolving): N/A
The serializers: no
The runtime per-record code paths (performance sensitive): unsure - TLS was already available, so if anyone is already using TLS, they are already experiencing some performance hit


statefun-e2e-tests/statefun-smoke-e2e-java/src/test/resources/remote-module/module.yaml
```diff
@@ -16,5 +16,11 @@
...
-  maxNumBatchRequests: 10000
\ No newline at end of file
+  urlPathTemplate: https://remote-function-host:8000
+  maxNumBatchRequests: 10000
+  transport:
+    type: io.statefun.transports.v1/async
+    trust_cacerts: file:/opt/statefun/modules/statefun-smoke-e2e/certs/a_ca.pem
+    client_cert: file:/opt/statefun/modules/statefun-smoke-e2e/certs/a_client.crt
+    client_key: file:/opt/statefun/modules/statefun-smoke-e2e/certs/a_client.key.p8
+    client_key_password: test
```
igalshilman, line 26: While I understand the need for a `client_key_password` to be present, it is a bit problematic, as it puts sensitive information into these yaml files.
How about we add an additional `client_key_password_file` property, and let people the option to reference that from a, say, k8s secret mounted?

statefun-e2e-tests/statefun-smoke-e2e-java/src/main/java/org/apache/flink/statefun/e2e/smoke/java/CommandInterpreterAppServer.java
```diff
@@ -20,33 +20,120 @@
...
+import org.apache.flink.shaded.netty4.io.netty.handler.ssl.SslContext;
+import org.apache.flink.shaded.netty4.io.netty.handler.ssl.SslContextBuilder;
+import org.apache.flink.shaded.netty4.io.netty.handler.ssl.SslProvider;
 import org.apache.flink.statefun.sdk.java.StatefulFunctionSpec;
 import org.apache.flink.statefun.sdk.java.StatefulFunctions;
-import org.apache.flink.statefun.sdk.java.handler.RequestReplyHandler;
+import org.apache.flink.statefun.sdk.java.slice.Slice;
+import org.apache.flink.statefun.sdk.java.slice.Slices;
 
 public class CommandInterpreterAppServer {
```
igalshilman, line 42: wow, well done!

igalshilman (commented):
Hi @FilKarnicki, the changes looks really good! really cool saying mTLS added to StateFun 🥳 

I've made a first pass over the PR and I have some comments, nothing fundamental.
Tomorrow I'll take a deeper look at the new integration test, but just wanted to provide some early feedback.

Thanks again for picking this up!

statefun-e2e-tests/statefun-smoke-e2e-java/src/test/resources/remote-module/module.yaml
FilKarnicki: I see, I'll make a change. I'm wondering if there's a way to read a pwd from a file while setting up TLS in Kafka. I'll have a poke around later :)

statefun-flink/statefun-flink-core/src/main/java/org/apache/flink/statefun/flink/core/nettyclient/NettyClient.java
```diff
@@ -208,4 +225,77 @@ private void releaseChannel0(Channel channel) {
...
+    Optional<String> maybeTrustCaCerts = spec.getTrustedCaCertsOptional();
+    Optional<String> maybeClientCerts = spec.getClientCertsOptional();
+    Optional<String> maybeClientKey = spec.getClientKeyOptional();
+    Optional<String> maybeKeyPassword = spec.getClientKeyPasswordOptional();
+
+    if (maybeClientCerts.isPresent() && !maybeClientKey.isPresent()) {
+      throw new IllegalStateException(
+          "You provided a client cert, but not a client key. Cannot continue.");
+    }
+    if (maybeClientKey.isPresent() && !maybeClientCerts.isPresent()) {
```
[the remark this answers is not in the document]
FilKarnicki: My intention was to write a kind of an XOR gate here. No cert/key is fine, both cert+key is fine, but not when there's one and not the other. Let me think if I can make this better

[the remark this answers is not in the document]
igalshilman: Oh, I see that make sense 👍

statefun-e2e-tests/statefun-smoke-e2e-java/src/test/resources/remote-module/module.yaml
igalshilman: I don't think I understand. If you can specify it in a `module.yaml` as a utf8 string, then it should be possible to read the exact same string from a file, no?

FilKarnicki:
hi @igalshilman I believe I addressed your concerns. Please let me know if there's anything else. Thanks! Fil

igalshilman (approved):
Thanks for addressing my comments @FilKarnicki, and for working on this issue 👍 
The changes look very very good to me, and the tests as well, I'll proceed to merging this PR.

PR #323, opened 2022-12-12
Feature/tb

tbal999:
this fixes the error:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:3.0.0:shade (default) on project statefun-smoke-e2e-embedded: Error creating shaded jar: Invalid signature file digest for Manifest main attributes -> [Help 1]
```

by removing the shaded jar plugin for 2 of the modules where this error occurred

downside:
- rather than 1 large jar, you get lots of little jars for these 2 modules

upside:
- we can do modifications to the modules without test failures

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,10 @@ This section contains information for building this project.
   $ mvn clean install -Prun-e2e-tests
   ```
 
+I had to use the below to successfully build the project:
+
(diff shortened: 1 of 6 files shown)


tbal999:
was pushed to wrong branch - ignore

PR #333, opened 2023-08-31
More retryable http status codes

gfmio:
This PR increases the list of status codes considered retryable.

* When Flink Statefun has high load, it can happen that the requests are sent too slowly and the server read timeout cuts off a request, in which case commonly a 400 error will be returned since the body was not parseable as a protobuf message.
* When using a load balancer like Envoy, 503 and 504 errors can occur frequently when the upstream service is unavailable or e.g. getting redeployed.


gfmio:
Question for the maintainers: Do you think we should keep this a fixed list or shall we make this configurable as a comma separated list in either the flink config or the `module.yaml`?

PR #335, opened 2023-08-31
enhance exponential backoff

gfmio:
This PR enhances the exponential backoff by adding jitter to prevent a thundering herd of requests.

The base backoff is increased to 1s, because 10ms can still create significant load on a service that might already be overwhelmed.

Finally, log messages are added or updated to provide useful diagnostic information in failure scenarios.


gfmio:
Question for the maintainers: Should we make the backoff settings configurable via the `module.yaml`?

PR #336, opened 2023-08-31
add failed offset to Flink error messages

gfmio:
This PR adds the topic, partition and offset to the error message when Flink encounters a null key to simplify finding the offending message.

### DIFF
--- a/statefun-flink/statefun-flink-io-bundle/src/main/java/org/apache/flink/statefun/flink/io/kafka/binders/ingress/v1/RoutableKafkaIngressDeserializer.java
+++ b/statefun-flink/statefun-flink-io-bundle/src/main/java/org/apache/flink/statefun/flink/io/kafka/binders/ingress/v1/RoutableKafkaIngressDeserializer.java
@@ -45,7 +45,7 @@ public final class RoutableKafkaIngressDeserializer
   public Message deserialize(ConsumerRecord<byte[], byte[]> input) {
     final String topic = input.topic();
     final byte[] payload = input.value();
-    final byte[] key = requireNonNullKey(input.key());
+    final byte[] key = requireNonNullKey(input);
     final String id = new String(key, StandardCharsets.UTF_8);
 
     final RoutingConfig routingConfig = routingConfigs.get(topic);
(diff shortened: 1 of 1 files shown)


statefun-flink/statefun-flink-io-bundle/src/main/java/org/apache/flink/statefun/flink/io/kafka/binders/ingress/v1/RoutableKafkaIngressDeserializer.java
```diff
@@ -45,7 +45,7 @@ public RoutableKafkaIngressDeserializer(Map<String, RoutingConfig> routingConfig
   public Message deserialize(ConsumerRecord<byte[], byte[]> input) {
     final String topic = input.topic();
     final byte[] payload = input.value();
-    final byte[] key = requireNonNullKey(input.key());
```
oleksandr-kazimirov, line 48: Hi, I'd like your thoughts on the idea of incorporating the ability to skip records with null keys. Additionally, I'm thinking of making this functionality configurable. What are your thoughts on this?
```
kind: io.statefun.kafka.v1/ingress
spec:
  id: localhost:9092
  consumerGroupId: test-group-dev
  startupPosition:
    type: earliest
  properties:
    - isolation.level: read_committed
    - security.protocol: SSL
    - ssl.truststore.location: /tmp/flink/jks/truststore.jks
    - ssl.truststore.password: changeit
    - ssl.keystore.location: /tmp/flink/jks/kafka_keystore.jks
    - ssl.keystore.password: test
    - ssl.key.password: test
  topics:
    - topic: test-dev
      skipNullKeyRecords: true
      valueType: ua.Test
      targets:
        - ua.Test
```
default can be skipNullKeyRecords: false

gfmio: That might work. The `NoOpDeserliazer` just returns `null`, which could be done here as well. https://github.com/apache/flink-statefun/blob/accd75ea0109845c4b4c0ddd74021147af1439d4/statefun-kafka-io/src/test/java/org/apache/flink/statefun/sdk/kafka/KafkaIngressBuilderTest.java#L166

PR #337, opened 2023-09-01
Max Retries

gfmio:
This PR adds a `maxRetries` option that limits the amount of times a request will be attempted before the request is dropped. When set to the default value of -1, batches will be retried indefinitely which matches the current behaviour.

This provides a simple means to drop failed messages after a configurable number of attempts when such behaviour is desired, which is currently not very easy to implement.

This also addresses an issue with Flink Statefun where a pipeline will get stuck if a remote function fails deterministically and the issue cannot be caught inside the handler. Normally, such issues can and should be caught inside the handler code, but this is not possible if the server deterministically crashes, e.g. with `log.Fatalw` in golang. Setting a maxRetries limit prevents such behaviour from blocking pipelines indefinitely.

### DIFF
--- a/statefun-e2e-tests/statefun-smoke-e2e-golang/src/test/resources/remote-module/module.yaml
+++ b/statefun-e2e-tests/statefun-smoke-e2e-golang/src/test/resources/remote-module/module.yaml
@@ -18,3 +18,4 @@ spec:
   functions: statefun.smoke.e2e/command-interpreter-fn
   urlPathTemplate: https://remote-function-host
   maxNumBatchRequests: 10000
+  maxRetries: 5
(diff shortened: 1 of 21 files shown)


davidradl:
Is there a Jira associated with this PR? The reason I ask, is that it would be useful to understand the impact of setting a max-retry and it failing in terms of the state of the system and recovery.  I suspect this would be worthy of explaining and getting feedback in a Jira.

gaoyibin0001:
There is already a issue here. 
https://issues.apache.org/jira/browse/FLINK-26424?jql=project%20%3D%20FLINK%20AND%20text%20~%20%22statefun%22

Seems this project  is no longer maintained.

PR #339, opened 2023-09-12
[FLINK-33075 ] Update outdated NOTICE files for Statefun

MartijnVisser:
This PR does 3 things:

1. Updated the copyright year
2. Sorts the existing entries in the NOTICE files
3. Updates the NOTICE files to match the actually bundled dependencies. This means that some entries are removed, because they are actually excluded, some version numbers were updated and some entries were added because they were missing in the NOTICE file, yet they were bundled.


dannycranmer (changes requested):

dannycranmer (approved):
LGTM, thanks

PR #341, opened 2023-09-26
[FLINK-33627] Bump snappy to 1.1.10.5

RyanSkraba:
## What is the purpose of the change

Bump the version of snappy to address a vulnerability (FLINK-33149)

### DIFF
--- a/statefun-kafka-io/pom.xml
+++ b/statefun-kafka-io/pom.xml
@@ -43,9 +43,7 @@ under the License.
             <version>${kafka.version}</version>
             <exclusions>
                 <!-- This collides with snappy-java brought from  
-                        org.apache.flink:flink-streaming-java_${scala.binary.version}
-                            org.xerial.snappy:snappy-java:1.1.4
-                -->
+                        org.apache.flink:flink-streaming-java -->
                 <exclusion>
                     <groupId>org.xerial.snappy</groupId>
                     <artifactId>snappy-java</artifactId>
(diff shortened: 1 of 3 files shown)


RyanSkraba:
I realised a bit late that this was a duplicate of dependabot's https://github.com/apache/flink-statefun/pull/340 -- there's some comment and unused property clean-up that could go along with this change.

statefun-kafka-io/pom.xml
```diff
@@ -43,9 +43,7 @@ under the License.
             <version>${kafka.version}</version>
             <exclusions>
                 <!-- This collides with snappy-java brought from  
-                        org.apache.flink:flink-streaming-java_${scala.binary.version}
-                            org.xerial.snappy:snappy-java:1.1.4
-                -->
+                        org.apache.flink:flink-streaming-java -->
```
davidradl, line 46: It would be good to leave a comment here. As far as I can see Flink 1.16.2 has snappy-java 1.1.8.3 which is vulnerable - so you want to exclude it here. But 1.17 Flink and above uses snappy-java 1.1.10.4. So this is a point in time change, because of your dependancy on the back level Flink. I assume we would want to move to a provided dependancy when we depend on a Flink 1.17 or above. Have I understood this correctly?

RyanSkraba: Hello!  If I understand correctly, the version of snappy brought in from flink-streaming-java *must* be compatible with the kafka client, or there's little hope of it working -- in my experience, this has always been the case with snappy patch releases, so 1.1.8.x should be OK with 1.1.10.x.

Users on Flink 1.16.2 will certainly have the vulnerability in flink-statefun (but also in all of the flink core APIs).  We're currently [voting on a 1.16.3 release](https://lists.apache.org/thread/dxfmt3v5n0xv5r9tjl30ob5d7y5t7pw3) with the bump. 

I'm open to a comment suggestion, but I'm not sure what would be useful or remain timely!  Wrangling dependencies is not an easy problem, so when I see an exclusion like this, I just assume the original author wanted a single authoritive source for the version.

